PR #301, opened 2011-06-15
Hooks directory

Spaceghost:
This should fix the lack of hooks/ directory under rvm. I know users can create it, but it saves someone going to IRC or some other place.


Boesmanbekkie:
Duplicate of #

Jsydog222:
Here is the property that I would like to buy

PR #1252, opened 2012-10-22
Allow wrapping multiple commands in one script

phoolish:
`rvm wrapper 1.9.2@me me "bundle exec unicorn_rails"` Creates 3 wrappers me_(bundle, exec, unicorn_rails)

`rvm wrapper 1.9.2@me me "bundle exec unicorn_rail"` Now creates me_bundle_exec_unicorn_rails

My bash string manipulation is a little rusty, so I had to break out the string replacement into its own line.

### DIFF
--- a/scripts/wrapper
+++ b/scripts/wrapper
@@ -76,8 +76,10 @@ symlink_binary()
   if
     wrap_binary && [[ -f "$file_name" ]]
   then
+    file_link="$rvm_bin_path/${prefix}_${binary_name##*\/}"
+    file_link=${file_link// /_}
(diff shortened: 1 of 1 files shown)


albb0920:
Just to point our this feature no longer works, now rvm will spit out 
"Could not find file 'bundle exec unicorn_rails' nor '/usr/local/rvm/wrappers/doujin/bundle_exec_unicorn_rails'"

I was trying to make /contrib/unicorn_init.sh to run unicorn_rails via bundler exec,
setting UNICORN="bundle exec unicorn_rails" won't work either because the init script quotes the whole thing.

PR #1764, opened 2013-04-06
add rvm osx-ssl-certs

mpapis:
idea from https://github.com/mpapis/veewee-templates-updater/issues/1


mcandre:
I ran `rvm osx-ssl-certs update all`, but `gitlab user` still fails with `certificate verify failed`.

https://github.com/NARKOZ/gitlab/issues/116

Any tips?

PR #2360, opened 2013-12-04
support for Cygwin requirements

mpapis:
RVM latest (as of this post) installs on Cygwin64 perfectly fine, but fails to compile/install rubies as it cannot figure out requirements.

Any way to get around this?


mpapis:
@dajoen check https://bugs.ruby-lang.org/issues/9555#change-50948

PR #3251, opened 2015-01-05
dropped railsexpress backport-401c8bb as it has been merged to ruby_2_2

skaes:
can you please merge?

otherwise railsexpress head patches will fail against current ruby-2_2 head


mpapis:
thank you!

PR #3272, opened 2015-01-22
sudo path hint

leoj3n:
Adds a hint to use `rvmsudo` when `GEM_HOME` and `GEM_PATH` aren't set in `sudo` shell.

### DIFF
--- a/scripts/rvm
+++ b/scripts/rvm
@@ -214,6 +214,13 @@ then
 Warning: PATH set to RVM ruby but GEM_HOME and/or GEM_PATH not set, see:
     https://github.com/wayneeseguin/rvm/issues/3212
 " >&2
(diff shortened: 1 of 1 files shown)


scripts/rvm
```diff
@@ -214,6 +214,13 @@ then
 Warning: PATH set to RVM ruby but GEM_HOME and/or GEM_PATH not set, see:
     https://github.com/wayneeseguin/rvm/issues/3212
 " >&2
+        if
+          [[ $UID -eq 0 ]]
```
mpapis: try `[[ -n "${SUDO_USER:-}" ]]` to avoid warning when ran as root

mpapis:
thank you!

Issue #3276, opened 2015-01-29
Ubuntu multi-user installation not working (because of UID variable)


tiangolo:
The multi-user installation is not working in Ubuntu, because the script checks if it was run with sudo using ` UID == 0 `, but that doesn't work in Ubuntu 12.04 nor Ubuntu 14.04. 

You can check (at an Ubuntu machine) with:
```
$ echo $UID # should return a normal user ID number, as 1000
1000
$ sudo echo $UID # the script expects 0, but the result is the same as without sudo
1000
```

I also checked that under CentOS 6.5 and the behaviour is the same, so, probably the problem also occurs under CentOS 6.5.

I will submit now a pull request with the fix.

mpapis:
how about:
```bash
echo $UID
sudo -i echo "\$UID"
```

mpapis:
or even more accurate:
```bash
sudo bash -c 'echo $UID'
```

tiangolo:
I already submitted a pull request with a fix: #3277 

What you suggest does work, but it also executes another bash session (if I'm not wrong).

In my pull request I suggest using `id -u`.

Would you check it out again?

If, after checking that pull request, you still prefer `bash -c 'echo $UID'`, I can change the code of the pull request to use that. And may I ask, why do you prefer that better?

mpapis:
please provide the requested tests and https://github.com/wayneeseguin/rvm/pull/3277#issuecomment-72121311

tiangolo:
OK, now I'm ashamed. I'm really sorry for making you lose your time.

I had understood that the multi-user install was just the same as the single-user install, but prefixed with `sudo`.

I was executing 

`sudo \curl -sSL https://get.rvm.io | bash -s stable`

instead of

`sudo \curl -sSL https://get.rvm.io | sudo bash -s stable`

mpapis:
don't prefix `\curl` with `sudo` - the download process does not require `sudo` at all, it's just a network connection.

PR #3277, opened 2015-01-29
Change UID check to id -u

tiangolo:
This fixes issue #3276 

The multi-user installation is not working in Ubuntu, because the script checks if it was run with sudo using ` UID == 0 `, but that doesn't work in Ubuntu 12.04 nor Ubuntu 14.04. 

You can check (at an Ubuntu machine) with:
```
$ echo $UID # should return a normal user ID number, as 1000
1000
$ sudo echo $UID # the script expects 0, but the result is the same as without sudo
1000
```
I also checked that under CentOS 6.5, and the behaviour is the same, so, probably the problem also occurs under CentOS 6.5.

-------------------

The way to find out if the script was run by sudo is with `id -u`, you can check (at an Ubuntu or CentOS machine) with:
```
$ id -u # should return a normal user ID number, as 1000
1000
$ sudo id -u # the script could expect 0, and the result would actually be 0
0
```

### DIFF
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -523,7 +523,7 @@ rvm_install_default_settings()
     [[ -z "${rvm_path:-}" ]]
   then
     if
-      (( UID == 0 ))
+      (( $( id -u ) == 0 ))
     then
       rvm_user_install_flag=0
       rvm_prefix="/usr/local"


mpapis:
the test you did are not valid - rejecting, please describe your problem in separate ticket

tiangolo:
I wrote a separate ticket: issue #3276 
(I just added the description of that ticket in this pull request)

And by the way, I ran the multi-user installation successfuly with this fix under Ubuntu.

mpapis:
then please test the `sudo bash -c 'echo $UID'`

tiangolo:
What you suggest does work, but it also executes another bash session (if I'm not wrong), and that would execute all the initialization steps, including all startup scripts, etc.

The actual code is supposed to work as stated in every piece of documentation I can find, but sadly it doesn't. The `id command` works, and: 
> This is the counterpart to the $UID, $EUID, and $GROUPS internal Bash variables.

as stated in: http://tldp.org/LDP/abs/html/system.html#IDREF

If, after checking it, you still prefer bash -c 'echo $UID', I can change the code of the pull request to use that. And may I ask, why do you prefer that better?

mpapis:
well running `\curl -sSL https://get.rvm.io | sudo bash` - involves another bash session - can you provide more details, especially a trace of the fail you are seeing? run:
```bash
\curl -sSL https://get.rvm.io | sudo bash -s head --trace 2>&1 | tee big.log
```
and [gist](https://gist.github.com) `big.log`

tiangolo:
OK, I'm sorry: https://github.com/wayneeseguin/rvm/issues/3276#issuecomment-72124366

PR #3306, opened 2015-02-18
Ignore ~/.rvmrc and /etc/rvmrc when --path is specified

timraasveld:
The `--path` parameter is not respected when installing (./install) if rvm_path and/or rvm_bin_path is defined in `~/.rvmrc` or `/etc/rvmrc`.

Not sure if it's desirable to ignore the whole rvmrc; is the install script supposed to respect other parameters from rvmrc?


mpapis:
maybe we should move loading `.rvmrc` before parsing params?

PR #3307, opened 2015-02-18
Better support rbx under FreeBSD 10

rys:
And start to warn that FreeBSD 10 and up is what rvm supports best.

### DIFF
--- a/scripts/functions/requirements/freebsd
+++ b/scripts/functions/requirements/freebsd
@@ -117,7 +117,7 @@ requirements_freebsd_define()
       then
         requirements_check gmake llvm35
         rvm_warn "If rbx installation fails, build it with clang and make sure llvm-config is in your PATH"
-        rvm_warn "Pass --cc clang --cxx clang++ to rvm install to build with clang on FreeBSD"
+        rvm_warn "Pass \"--cc clang --cxx clang++\" to rvm install to build with clang on FreeBSD"
       else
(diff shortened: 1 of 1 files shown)


scripts/functions/requirements/freebsd
```diff
@@ -58,6 +58,12 @@ requirements_freebsd_before_detect_package_manager()
 
 requirements_freebsd_before()
 {
+  if
+    __rvm_version_compare ${_system_version} -le 10.0
```
mpapis: I think you meant `-lt` as reverse to `-ge` that shows up later

scripts/functions/requirements/freebsd
```diff
@@ -105,6 +111,17 @@ requirements_freebsd_define()
...
+      if
+        __rvm_version_compare ${_system_version} -ge 10.0
+      then
+        requirements_check gmake llvm35
+        rvm_warn "If rbx installation fails, build it with clang and make sure llvm-config is in your PATH"
+        rvm_warn "Pass \"--cc clang --cxx clang++\" to rvm install to build with clang on FreeBSD"
+      else
+        rvm_warn "rbx installation by rvm is only supported on FreeBSD 10 or higher"
+      fi
+      ;;
```
mpapis: https://github.com/rys/rvm/blob/freebsd-10/scripts/functions/requirements/ubuntu#L99-L108 - maybe just add the packages and set variables?

PR #3326, opened 2015-03-02
+ruby-2.2.1

freibuis:
2.2.1 listed in the cache http://cache.ruby-lang.org/pub/ruby/.. but it is not listed on the download page (https://www.ruby-lang.org/en/downloads/) yet!

### DIFF
--- a/config/known
+++ b/config/known
@@ -8,6 +8,7 @@
 [ruby-]2.1.4
 [ruby-]2.1[.5]
 [ruby-]2.2.0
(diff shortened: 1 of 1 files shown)


havenwood:
It took a few days but 2.2.1 is now listed: https://www.ruby-lang.org/en/news/2015/03/03/ruby-2-2-1-released/

I went ahead and started PR #3329 that includes the known versions bump and checksums. This should be closed by #3329.

Issue #3262, opened 2015-01-13
Error after `rvm get 1.26.9`


kronn:
I tried to update RVM from 1.25.15 to 1.26.9.

```
root@example.net # /usr/local/rvm/bin/rvm get 1.26.9

Hi there, it looks like you have requested updating rvm to version 1.26.9,
if your intention was ruby installation use instead: rvm install 1.26.9

Downloading https://get.rvm.io
Downloading https://github.com/wayneeseguin/rvm/archive/1.26.9.tar.gz
Downloading https://github.com/wayneeseguin/rvm/releases/download/1.26.9/1.26.9.tar.gz.asc
gpg: Signature made Mi 07 Jan 2015 12:55:10 CET using RSA key ID BF04FF17
gpg: Good signature from "Michal Papis (RVM signing) <mpapis@gmail.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 409B 6B17 96C2 7546 2A17  0311 3804 BB82 D39D C0E3
     Subkey fingerprint: 62C9 E5F4 DA30 0D94 AC36  166B E206 C29F BF04 FF17
GPG verified '/usr/local/rvm/archives/rvm-1.26.9.tgz'

Upgrading the RVM installation in /usr/local/rvm/
Upgrade of RVM in /usr/local/rvm/ is complete.

# xunde,
#
#   Thank you for using RVM!
#   We sincerely hope that RVM helps to make your life easier and more enjoyable!!!
#
# ~Wayne, Michal & team.

In case of problems: http://rvm.io/help and https://twitter.com/rvm_io

Upgrade Notes:

  * No new notes to display.

/usr/local/rvm/bin/rvm: line 67: unexpected EOF while looking for matching `"'
/usr/local/rvm/bin/rvm: line 70: syntax error: unexpected end of file
/usr/local/rvm/bin/rvm: line 1: return: can only `return' from a function or sourced script
```

The exitstatus is "2".

However, the new version is installed.

Sadly, our automation setup only sees the exitstatus and fails.

The new rvm-script has fewer lines than noted in the error message. therefore, I have problems debugging this myself.

kronn:
The username "xunde" is the name of the login-user

```
root@example.net # env | grep USER
USER=root
SUDO_USER=xunde
USERNAME=root
```

kronn:
I just checked on another machine having the same issue.

[...]

PR #3340, opened 2015-03-10
Fix trap behavior

killphi:
I have done as was asked in #3262, however I couldn't get https://github.com/wayneeseguin/rvm-test to work properly.

Also, I have no idea for how to actually and properly test this change.

I did however take the liberty to rename the signal numbers into their actual names, making it a bit more readable to the human eye.

### DIFF
--- a/scripts/functions/environment
+++ b/scripts/functions/environment
@@ -204,7 +204,7 @@ __rvm_setup()
 
   if [[ -n "${BASH_VERSION:-}" ]] && ! __function_on_stack cd pushd popd
   then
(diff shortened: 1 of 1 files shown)


mpapis:
thank you!

jaspertandy:
Happy to test this if needed. Let me know the rvm code to pull the appropriate version and I'll do so.

mpapis:
@jaspertandy:
```bash
rvm get head
rvm install ruby-named
```
and then CTRL+C somewhere, in another try, `killall rvm` - although you might need to find the proper shell process first and kill it manually

mpapis:
the most important part is that the shell result code should be `!=0` check with `echo $?`

PR #3364, opened 2015-04-01
Added some missing ${}s to install scripts

### LINKED ISSUE #3213: GPG problems on Cygwin?
I have installed GPG and it seems to work from my cygwin zsh terminal, however I get an error each time I attempt to install RVM.

Here is the output of `gpg --keyserver hkp://keys.gnupg.net --recv-keys D39DC0E3`:
```
gpg: requesting key D39DC0E3 from hkp server keys.gnupg.net
gpg: key D39DC0E3: "Michal Papis (RVM signing) <mpapis@gmail.com>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
```

Then when I attempt to install with `\curl -sSL https://get.rvm.io | bash -s stable`, I get:
```
Downloading https://github.com/wayneeseguin/rvm/archive/1.26.4.tar.gz
Downloading https://github.com/wayneeseguin/rvm/releases/download/1.26.4/1.26.4.tar.gz.asc
gpg: can't open `/home/username/.rvm/archives/rvm-1.26.4.tgz.asc': No such file or directory
gpg: verify signatures failed: No such file or directory
Warning, RVM 1.26.0 introduces signed releases and automated check of signatures when GPG software found.
Assuming you trust Michal Papis import the mpapis public key (downloading the signatures).

GPG signature verification failed for '/home/username/.rvm/archives/rvm-1.26.4.tgz' - 'https://github.com/wayneeseguin/rvm/releases/download/1.26.4/1.26.4.tar.gz.asc'!
try downloading the signatures:

    gpg2 --keyserver hkp://keys.gnupg.net --recv-keys D39DC0E3

or if it fails:

    command curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -

the key can be compared with:

    https://rvm.io/mpapis.asc
    https://keybase.io/mpapis
```

This file does appear to exist, but I keep hitting walls. Is there a flag that I can pass the rvm installation script to bypass this security check?

bezeek:
Should fix #3213.

Tested with the presence of `gpg2.exe` provided by Gpg4win and `gpg.exe` provided by Cygwin, with and without a stand-in `gpg2.exe` (just a copy of `gpg.exe`) in the Cygwin path. In each case, Gpg4win's version was ignored as desired, and rvm installation proceeded happily.

### DIFF
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -388,7 +388,7 @@ rvm_install_gpg_setup()
   export rvm_gpg_command
   {
     rvm_gpg_command="$( \which gpg2 2>/dev/null )" &&
-    [[ rvm_gpg_command != "/cygdrive/"* ]]
+    [[ ${rvm_gpg_command} != "/cygdrive/"* ]]
   } ||
     rvm_gpg_command="$( \which gpg 2>/dev/null )" ||
     rvm_gpg_command=""
--- a/scripts/functions/cli
+++ b/scripts/functions/cli
@@ -203,7 +203,7 @@ rvm_install_gpg_setup()
 {
   {
     rvm_gpg_command="$( \which gpg2 2>/dev/null )" &&
-    [[ rvm_gpg_command != "/cygdrive/"* ]]
+    [[ ${rvm_gpg_command} != "/cygdrive/"* ]]
   } ||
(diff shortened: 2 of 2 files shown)


mpapis:
thank you!

PR #3379, opened 2015-04-14
Code of conduct

rys:


mpapis:
@rys can you push the same to all other repos in rvm organization?

Issue #3395, opened 2015-04-25
Automatically set login shell option for gnome-terminal


Folis:
After some discussion with @mpapis we decided that it would be nice to automatically set the login shell option for gnome-terminal, so that the infamous "RVM is not a function" message does not appear. I figured out that the setting can be set here:

`~/.gconf/apps/gnome-terminal/profiles/Default/%gconf.xml`

Setting this option should do it:

`<entry name="login_shell" mtime="<unix_timestamp>" type="bool" value="true"/>`

I'd do it myself, but I don't know too much about shell scripting.

mpapis:
fixed !

PR #3396, opened 2015-04-25
Automatically set the login shell option for gnome-terminal

Folis:
This greps for the login-shell option in the Default profile of gnome-terminal and sets it, if not already existing. Thus you will never have to set it manually, avoiding the 'RVM is not a function' message

Solves issue #3395


mpapis:
thank you!

PR #3406, opened 2015-05-04
Take the SECRET_KEY_BASE env variable used by Ruby on Rails

StephaneJuban:
In production environment, Ruby on Rails used the environment variable SECRET_KEY_BASE but it is not pass without this change.

The complete issue can be found on [Stackoverflow](http://stackoverflow.com/questions/30000303/missing-secret-key-base-in-rails-production), I did some test and it's OK for me after this quick change.

### DIFF
--- a/bin/rvmsudo
+++ b/bin/rvmsudo
@@ -81,7 +81,7 @@ then
   eval command sudo \"\${sudo_args[@]}\" $(
     /usr/bin/env |
       awk -F= 'BEGIN{v=0;}/^[a-zA-Z_][a-zA-Z0-9_]*=/{v=1;}v==1&&$2~/^['\''\$]/{v=2;} v==1&&$2~/^\(\) \{/{v=0;} v==1&&$2~/^\(/{v=3;}v==2&&/'\''$/&&!/'\'\''$/{v=1;}v==3&&/\)$/{v=1;}v{print;}v==1{v=0;}' |
-      GREP_OPTIONS="" \grep -E '^rvm|^gemset|^http_|^PATH|^IRBRC|RUBY|GEM' |
+      GREP_OPTIONS="" \grep -E '^rvm|^gemset|^http_|^PATH|^IRBRC|^SECRET_KEY|RUBY|GEM' |
       sed 's/=\(.*\)$/="\1"/'
   ) \"\$@\" ||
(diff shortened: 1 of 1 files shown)


StephaneJuban:
What do you think about it @mpapis ?

mpapis:
I'm confused, what is the full command that was failing? why would you use `rvmsudo` for starting your production?

phillipp:
Besides, I'm not happy with pushing all ENV that begins with SECRET to the spawned process. In fact I'm not a fan of managing secrets by environment (because every subprocess has access to it and may for example dump it out in error pages etc), but that's another case.

I would prefer to only push the SECRET_KEY_BASE to rails and not all SECRET-prefixed ENV vars, because that is not specific enough. You may have other ENV vars prefix with SECRET that you don't want to export to the process.

StephaneJuban:
@mpapis The complete command I used to start my server is `rvmsudo rails s -e "production" -p 80 -d`, I have always done that but it's may not be the right way...

@phillipp Yes I'm also not a big fan of managing secrets by env variables but I did this pull request first of all because I encountered the issue and secondly because [Rails 4.1 introduced it by default](https://github.com/rails/rails/commit/b9095ec95c38e17425eef26018aeb962e96c323f). We can add the complete sentence for grep `SECRET_KEY_BASE` this was my first tought

StephaneJuban:
Just updated *SECRET_KEY-prefixed* grep to full **SECRET_KEY_BASE**

mpapis:
@StephaneJuban this is the slowest way to run rails, try this:

1. use `nginx` or `apache` as front of your application handing port `80`
2. use `passenger`, `unicorn`, `mongrel` or anything like it to handle your `rails` application

StephaneJuban:
Ok @mpapis I will update my method to run my Rails app.
So what about this pull request ? Is it unecessary in your opinion ?

mpapis:
Yes I would avoid such changes which are dictated by something nobody should use, please close if you agree - or prove your point if you think it makes sense.

StephaneJuban:
We can close this one. [As you pointed out](https://github.com/rvm/rvm/pull/3406#issuecomment-99825051) I will try an other way to run my rails app. Thanks !

PR #3452, opened 2015-06-26
cygwin: package names for mingw64-*-gcc changed

mogria:
The packages `mingw64-i686-gcc` and `mingw64-x86_64-gcc` got renamed and received a `-core` extension
https://cygwin.com/cgi-bin2/package-grep.cgi?grep=mingw64-.*-gcc&arch=x86_64


havenwood:
Thanks!

PR #3468, opened 2015-07-16
Container-based travis builds

donv:
The container-based builds should be faster than the VM-based builds.

http://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: c
 os: [linux, osx]
 osx_image: mavericks
+sudo: false
+addons:
+  apt:
+    packages:
+    - gawk
 before_install:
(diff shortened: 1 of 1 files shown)


donv:
Ooops, one extra commit was included.

donv:
OK, removed the extra commit.

.travis.yml
```diff
@@ -1,6 +1,11 @@
 language: c
 os: [linux, osx]
 osx_image: mavericks
+sudo: false
+addons:
+  apt:
+    packages:
+    - gawk
```
havenwood: This appears to be necessary for some reason at this point but may be able to be removed later.

havenwood:
Thanks!

PR #3469, opened 2015-07-18
SSL/TLSize all the things! (convert http:// and git:// to https:// where possible)

reedloden:
Even though RVM uses hardcoded checksums to help prevent MITM and corruption issues, using SSL/TLS to download the files adds another layer of protection, ensuring that nobody can see that somebody is downloading a certain ruby version.

Also, another reason to use https:// is that it doesn't seem like every file downloaded has a SHA512 hash (only MD5 hash), and MD5 is known to have numerous weaknesses, so trusting it alone + http:// == danger!


scripts/functions/pkg
```diff
@@ -279,8 +279,8 @@ install_openssl()
 update_openssl_certs()
 (
   __rvm_cd "$rvm_usr_path/ssl" &&
-  command curl -O http://curl.haxx.se/ca/cacert.pem &&
-  \command \mv cacert.pem cert.pem
+  command curl -O https://raw.githubusercontent.com/bagder/ca-bundle/master/ca-bundle.crt &&
```
richo: It's a little unclear that this is better?

richo:
Looks reasonable to me, but I'll let @mpapis have the final say

scripts/functions/pkg
reedloden: Can you explain? The current version uses http:// to download a list of CA roots with no type of authenticity checking at all, making it easy for somebody to MITM that URL to add a malicious CA root. I just replace it with an official https:// version (as per http://curl.haxx.se/docs/caextract.html). Of course, there is a slight chicken and egg problem here, as by using https:// to download the CA roots, you are requiring that `curl` itself has a proper CA root repository enough to download something from GitHub. However, considering the link to download & install RVM is over https:// as well, I think that's acceptable. Sadly, there's no GPG-signed version of this file that we could use for validation (similar to what https://rvm.io/rvm/security does, though that page recommends you download a GPG public key insecurely, sigh). Considering how frequently RVM is updated and released, another option is to actually include the CA bundle itself as part of RVM. However, not sure that's wanted... Any further comments/thoughts are appreciated.

scripts/tools
```diff
@@ -176,12 +176,12 @@ tools_mirror()
     if
      (( ${rvm_force_flag:-0} == 1 ))
     then
-      __rvm_sed_i "${file}" -e "s/^ruby_${n}_url=.*$/ruby_${n}_url=http:\/\/www.mirrorservice.org\/sites\/ftp.ruby-lang.org\/pub\/ruby\/${n}/"
+      __rvm_sed_i "${file}" -e "s/^ruby_${n}_url=.*$/ruby_${n}_url=https:\/\/www.mirrorservice.org\/sites\/ftp.ruby-lang.org\/pub\/ruby\/${n}/"
```
reedloden: Note that https://www.mirrorservice.org/sites/ftp.ruby-lang.org/pub/ruby doesn't exist anymore, so I don't know how useful this code is in the first place.

config/db
```diff
@@ -3,10 +3,9 @@ niceness=0
 # Rubies
 interpreter=ruby
 default_ruby=ruby
-ruby_repo_url=git://github.com/ruby/ruby.git
-ruby_url=http://cache.ruby-lang.org/pub/ruby
-ruby_url_fallback_1=http://ftp.ruby-lang.org/pub/ruby
-ruby_url_fallback_2=http://www.mirrorservice.org/sites/ftp.ruby-lang.org/pub/ruby
```
reedloden: Removed, as this is 404.

scripts/functions/pkg
richo: Oh right, It was unclear that there was a link between that github user and
the curl upstream (Sorry on planeternets).

My gut feel is that vendoring it is probably least wrong.

reedloden: Ah, ok. Yeah, @bagder is the `curl` upstream maintainer (Daniel Stenberg).

config/db
```diff
@@ -50,36 +49,36 @@ rbx_1.2.3_patch_level=20110315
 rbx_1.2.4_patch_level=20110705
 rbx_url=http://asset.rubini.us
 rbx_2.0.0_url=http://releases.rubini.us
```
reedloden: I'd love to change this to https://s3.amazonaws.com/releases.rubini.us (since releases.rubini.us just points to the S3 bucket anyway). However, wasn't sure if that was considered kosher.

config/db
```diff
@@ -50,36 +49,36 @@ rbx_1.2.3_patch_level=20110315
...
-rubinius_repo_url=git://github.com/rubinius/rubinius.git
+rubinius_repo_url=https://github.com/rubinius/rubinius.git
 ree_version=1.8.7
 ree_configure_flags=--dont-install-useful-gems
 ree_1.8.6_patch_level=20090610
 ree_1.8.6_url=http://rubyforge.org/frs/download.php/58677
```
reedloden: This doesn't exist anymore. The only working place I could find is http://mirror.as24220.net/pub/rubygems/files/emm-ruby/ (sadly, https:// not supported).

config/db
```diff
@@ -21,10 +20,10 @@ CentOS_5_ruby_1.9.2_patch_level=p320
 CentOS_5_ruby_1.9.2_patch_level_warn=Warning, Centos 5 can not build head version of ruby, falling back to latest patchlevel, your ruby will be less secure bacuase of this.
 CentOS_5_ruby_1.8.7_patch_level=p374
 CentOS_5_ruby_1.8.7_patch_level_warn=Warning, Centos 5 can not build head version of ruby, falling back to latest patchlevel, your ruby will be less secure bacuase of this.
-rubygems_url=http://production.cf.rubygems.org/rubygems
+rubygems_url=https://rubygems.global.ssl.fastly.net/rubygems
```
reedloden: This is what the rubygems.org folk recommended when I asked about SSL for the CloudFront instance (which is what production.cf.rubygems.org uses). Also, they are deprecating CloudFront soon anyway (replacing it with Fastly completely, though maybe something else in the future).

bagder:
Let me just confirm that I maintain both the site (curl.haxx.se), the scripts that converts the Mozilla ca-certs as well as the github mirror of the ca-bundle at https://github.com/bagder/ca-bundle. Enjoy.

cvss:
Switching to HTTPS from HTTP prevents using of local caching http proxies. It's a big disadvantage, because http caches located in the office or data-center speed up all downloads 10x-100x times.

scripts/functions/pkg
mpapis: `+1` for `https`
`-1` for this change
maybe something like this:
```bash
__rvm_curl https://raw.githubusercontent.com/bagder/ca-bundle/master/ca-bundle.crt -o cert.pem ||
__rvm_curl http://curl.haxx.se/ca/cacert.pem -o cert.pem
```

reedloden: That's opening up a security problem for people, as somebody who is actively MITM'ing could block access to raw.githubusercontent.com, which would cause the fallback to http://, compromising the user's CA roots. Curious why you would want to support http:// at all, when to download rvm, the all the recommendations use https://.

I think it's fine to document something as an alternative (maybe if curl fails, throw a warning mentioning it and stating that the user could get the file manually, with the caveat of the MITM risk), but automated insecure fallback doesn't help the user. Such fallbacks have been actively used to exploit problems in SSLv3 or other protocols, so it's not something that's unheard of.

sikachu:
I would love to hear how @mpapis think about this too.

However, for me, I think it's best for us to offer the security up to a certain point. We really should try to ensure that the Ruby that user installed via RVM is not being tainted by some third party.

So, I think my take would be:

* I agree with the change that make us download certificate bundle via HTTPS. This _must_ not be changed by MITM, otherwise we are so screwed.
* I disagree with the change that make us download all packages via HTTP. Some people rely on a local proxy server to cache those files. RVM already has a mechanism to validate the packages against known md5/sha512, so we should be good. If packages are missing from the list, please help us add it (with citation on the PR about where you got it from).
* I agree with changing all appropriate web page links from HTTP to HTTPS. We have no control over validating the information, so sending user through a _secure_ channel is always better if website supports it.

With that said, I would love to see a PR that adds a feature or flag that force all download to go through HTTPS. This will be very beneficial for those users who does not have bandwidth constraint and would love an enhanced security.

scripts/functions/pkg
mpapis: the problem is that we are downloading certificates because the existing ones fail, there is nothing that would protect you from MITM when your certificate for the given site is not valid

mpapis:
I'm ok with SSL, with the exception of https://github.com/rvm/rvm/pull/3469#discussion-diff-34960307

reedloden:
@mpapis I will update the diff, keeping https:// everywhere else except that one spot.

reedloden:
@mpapis (and others) -- this has been updated, and Travis CI build is green.

reedloden:
It's been 6+ months since I submitted this, and I've had to continually rebase this regularly. Just curious what's blocking this from getting reviewed and merged? Would really like to get this into rvm. @mpapis @sikachu 

As far as I'm aware, rvm is the only ruby build manager that is still using http:// for downloads. Everybody else (including upstream ruby) has moved to https://.

Let me know if there's anything else I can do to help.

richo:
Sorry for the delay here, thanks for the patch!

Issue #3465, opened 2015-07-15
RVM doesn't operate as expected


smuth4:
From the documentation (https://rvm.io/deployment/cron), I expect `rvm env --path -- 1.9.3` to give me the path to the environment file for 1.9.3 regardless of the actual ruby version being used. From my experiments, and what I've seen of the shell script source, it will always return the environment file for the currently used ruby verision. It's possible I'm just misreading the documentation, but if that's true, that might want to be made more explicit.

Environment:
```
$ rvm -v
rvm 1.26.11 (stable) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]
$ rvm use system
Now using system ruby.
$ ruby -v
ruby 1.8.7 (2011-06-30 patchlevel 352) [x86_64-linux]
$ rvm list

rvm rubies

   ruby-1.9.3-p551 [ x86_64 ]

# Default ruby not set. Try 'rvm alias create default <ruby>'.

# => - current
# =* - current && default
#  * - default
$ ls -l /usr/local/rvm/environments
total 0
lrwxrwxrwx 1 root rvm 12 Jul 15 17:04 environment -> /environment
lrwxrwxrwx 1 root rvm 47 Jul 15 15:36 ruby-1.9.3-p551 -> /usr/local/rvm/gems/ruby-1.9.3-p551/environment
lrwxrwxrwx 1 root rvm 54 Jul 15 15:36 ruby-1.9.3-p551@global -> /usr/local/rvm/gems/ruby-1.9.3-p551@global/environment
```

What happens:
```
$ rvm env --path
/usr/local/rvm/environments/system
$ rvm env --path 1.9.3
/usr/local/rvm/environments/system
$ rvm env --path 1.9.3-p551
/usr/local/rvm/environments/system
$ rvm env --path 30000.0.0
/usr/local/rvm/environments/system
$ rvm env --path garbage
Unrecognized command line argument: 'garbage' ( see: 'rvm usage' )
```

What I expect to happen:
```
$ rvm env --path
/usr/local/rvm/environments/system
$ rvm env --path 1.9.3
/usr/local/rvm/environments/ruby-1.9.3-p551
$ rvm env --path 1.9.3-p551
/usr/local/rvm/environments/ruby-1.9.3-p551
$ rvm env --path 30000.0.0
# Some sort of "ruby not found" error
$ rvm env --path garbage
[...]

PR #3483, opened 2015-08-08
Output correct environment path.

bronzdoc:
Pull requets to fix the  "'env' command doesn't operate as expected #3465" issue

### DIFF
--- a/scripts/cli
+++ b/scripts/cli
@@ -656,6 +656,7 @@ __rvm_parse_args()
             rvm_token=${rvm_token#--}
             rvm_token=${rvm_token//-/_}
             export "rvm_${rvm_token}_flag"=1
+            rvm_ruby_args+=($next_token);
             ;;
 
           --no-docs)
--- a/scripts/functions/env
+++ b/scripts/functions/env
@@ -272,7 +272,26 @@ __rvm_env_print()
   if
     [[ "$rvm_path_flag" == "1" || "$*" == *"--path"* ]]
   then
-    echo "$environment_file_path"
+    if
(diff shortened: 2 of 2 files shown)


bronzdoc:
The build is failing here: 

<code>rvm env 2.1.1 -- --path</code>
rvm: Environment --path not found.
# failed: match = /2.1.1/
# failed: match = /environments/

i noticed that the way the command is being tested is not as the one explained in the documentation of rvm. https://rvm.io/deployment/cron#loading-rvm-environment-files-in-shell-script
 
to fix this i need to know what is the correct way to execute the command, thanks.

scripts/functions/env
```diff
@@ -272,7 +272,26 @@ __rvm_env_print()
   if
     [[ "$rvm_path_flag" == "1" || "$*" == *"--path"* ]]
   then
-    echo "$environment_file_path"
+    if
+      [[ ! -z "$@" ]]
```
havenwood: `-n` is a tad more concise than `! -z` for the string-isn't-null test.

bronzdoc: you are right @havenwood thanks, i will update it.

scripts/cli
```diff
@@ -656,6 +656,7 @@ __rvm_parse_args()
             rvm_token=${rvm_token#--}
             rvm_token=${rvm_token//-/_}
             export "rvm_${rvm_token}_flag"=1
+            rvm_ruby_args+=($next_token);
```
mpapis: this does not look good, the options here do not take arguments and `next_token` is what's listed after them, could be anything

PR #3503, opened 2015-09-08
Making alias's error message reflect actual usage.

acook:
Alias was giving an improper syntax in it's error message, it only indicated 2 argument, but it requires 3.


franklinyu:
This should be merged.

havenwood:
Thank you!

Issue #623, opened 2011-11-29
Compiling with 32bit flags leads to 64bit ruby


apeiros:
Hi

I'm using Kenneth Reitz' GCC installer (https://github.com/kennethreitz/osx-gcc-installer), the newest rvm (1.10.0-pre) and OS X Lion 10.7.2.
I tried to install ruby 1.9.2p290 as 32bit executable, following http://beginrescueend.com/support/troubleshooting/ (section "i386 (32 bit)").
However, the compiled executable is 64bit.

Additional information:

```
$ uname -mprsv
Darwin 11.2.0 Darwin Kernel Version 11.2.0: Tue Aug  9 20:54:00 PDT 2011; root:xnu-1699.24.8~1/RELEASE_X86_64 x86_64 i386

$ rvm_archflags="-arch i386" CFLAGS='-m32' CXXFLAGS='-m32' LDFLAGS='-m32' rvm install ruby-1.9.2
# …snip…
Install of ruby-1.9.2-p290 - #complete 
$ rvm use 1.9.2
$ which ruby
/Users/apeiros/.rvm/rubies/ruby-1.9.2-p290/bin/ruby
$ file /Users/apeiros/.rvm/rubies/ruby-1.9.2-p290/bin/ruby
/Users/apeiros/.rvm/rubies/ruby-1.9.2-p290/bin/ruby: Mach-O 64-bit executable x86_64
$ cat ~/.rvm/src/ruby-1.9.2-p290/Makefile | grep '^arch ='
arch = x86_64-darwin11.2.0
```

If I can supply any other information, please let me know.

eevee:
Almost four years later, I wasted half a day chasing down a similar problem, because the flags are still not documented.  I finally found them reading the source code.

v6:
// , Interesting that rubygems seems to use the system architecture for building native extensions. 

@eevee, how did you find this bug?

eevee:
I have a 64-bit Linode running a 32-bit kernel.

It was originally all 32-bit, then they gave me a free upgrade to a 64-bit host, and I didn't see any good reason to redo a lot of ad-hoc stuff just to fix the mismatch.

PR #3526, opened 2015-10-11
Update README with --32 bit flag

RochesterinNYC:
32 bit flag exists but is not documented yet as per this issue: https://github.com/rvm/rvm/issues/623


RochesterinNYC:
Hey @havenwood, any chance you could check this out? It's a really quick change but think it'd be a really useful thing to have in the docs.

havenwood:
Thanks!

PR #3527, opened 2015-10-14
change jruby release version info to 9.0.1.0 on `rvm list known`

southwolf:


havenwood:
Thank you! Closing only because this was bumped straight to 9.0.3.0 in #3534 before we merged your commit.

PR #3528, opened 2015-10-19
Allow $rvm_path to contain a space

steven-aerts:
An rvm_path which contains a space is currently causing an error.


sikachu:
This looks like a good change.

However, before I merge it in, would you mind going through the file and see if there are more places that we should quote around `$rvm_path`? I feel like there could be more ...

mpapis:
there is a lot more, at least there was, that's why we had a warning if there is a space do trick for it: https://github.com/steven-aerts/rvm/blob/patch-1/binscripts/rvm-installer#L751-L763

PR #3538, opened 2015-10-26
Added checksum of RubyGems 2.4.8

emmanne08:
* Fix for the following error when installing a Ruby with the flag `--verify-downloads 0`

    There is no checksum for 'http://production.cf.rubygems.org/rubygems/rubygems-2.4.8.tgz' or 'rubygems-2.4.8.tgz', it's not possible to validate it.
    This could be because your RVM install's list of versions is out of date. You may want to
    update your list of rubies by running 'rvm get stable' and try again.
    If that does not resolve the issue and you wish to continue with unverified download
    add '--verify-downloads 1' after the command.
    
    There has been an error while trying to fetch rubygems.
    Halting the installation.


havenwood:
Thanks!

PR #3547, opened 2015-11-06
Update README to explicitly indicate --64 flag

RochesterinNYC:
This is a change that goes along with https://github.com/rvm/rvm/pull/3526.

Adds the --64-bit flag to the README which is the option for installing 64-bit rubies (although this is also default behavior).


sikachu:
Thanks for the patch!

PR #3550, opened 2015-11-13
Detect an absence of grep

ueokande:
When grep is not installed, RVM installer script misdetects that tar is not installed even if tar is installed.


sikachu:
Looks good. Thanks for the patch!

PR #3601, opened 2016-01-09
Add rbx-3, drop old/broken versions

ojab:


henderea:
+1
I was thinking about playing around with rubinius, but with RVM not supporting the 3.x versions, it's a little hard for me to set it up.

I use rvm head, so all I really need is for this to be merged to rvm:master.  Hopefully that can happen soon.

havenwood:
The new versions look good but I'm not sure if it's best to keep the old Rubinius version support?

@mpapis Keep or delete no-longer-supported Rubinius version metadata?

mpapis:
not supported does not mean no one wants to install it, few days ago I just had request for REE - It's still rolling somewhere - usually the policy was to leave the old versions working as long as we were able to hack that

ojab:
Why someone will need new rvm for rbx-1, for example, if rbx-1 cannot be built on any up-to-date platform and why rubies that already broken for a long time without any mentions in issues can't be removed?

I can fix rbx-1 installation by replacing broken links w/ link to releases on github, but I can't even test something except download success without installing archeological linux distributions and reading README for each release to find matching llvm/etc versions.
I don't see any sense in this either.

mpapis:
adding the new versions should not require changes in `selector_interpreters`, maybe except renaming `rbx_2.0.0_url` to `rbx_new_url`

mpapis:
@ojab you can not install it does not mean it wont be something others will want to do, it requires no effort to keep it in RVM, removing it only here is pointless, there is a lot more code that made it work, if you are cleaning - remove it all or just leave it be (for others to use if they wish so)

AquarHEAD:
@mpapis @ojab Can't we first add new versions, and open another PR for removing old ones if that's really needed? Want to play with rbx-3 recently as well...

mpapis:
@AquarHEAD open a PR with adding only and we can merge it right away.

pkuczynski:
I just added latest rubinus 3.59 so you can play around with it. I am closing this PR as we would like to keep the old rubinius versions for the time being.

mxup:
@pkuczynski 
So there is no reason in working with / on rbx 3.1 any longer?
I was kinda drawn to rbx-3[.107], me having fairly old gems

Is 2.3 a way to go here? 
I'm fairly new with versioning but I feel the beauty when I'm to stumbling upon it.

Issue #3560, opened 2015-11-17
OSX (El Capitan) rvm source in .bash_profile ruins .bash_history writing (Terminal.app)


tilhoft:
After I upgraded to OSX El Capitan, the .bash_history is not writing, if this line is in my  .bash_profile
[[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm"

tilhoft:
Turned out the new feature called bash_sessions ruins this.
temporary solution(/workaround) to disable it: http://stackoverflow.com/questions/32418438/how-can-i-disable-bash-sessions-in-os-x-el-capitan

Yanchek99:
Relates to #3540

tilhoft:
Any improvements on this? macOS Sierra still involved?

PR #3627, opened 2016-02-15
Bug with bash sessions on OSX 10.11+

### LINKED ISSUE #3540: RVM overwrites bash exit trap defined in OS X El Capitan
OS X El Capitan introduced sessions to manage history. History is written to files in ~/.bash_sessions/ instead of ~/.bash_history when bash exits. This is driven by an exit trap that calls shell_session_update(). RVM overwrites this trap, meaning that history isn't saved when the bash session is exited after running RVM.

This occurs with the RVM called in ~/.bashrc:

    $ trap
    trap -- 'shell_session_update' EXIT
    $ source .rvm/scripts/rvm
    $ trap
    $

And with the regular RVM:

    $ trap
    trap -- 'shell_session_update' EXIT
    $ rvm --version
    rvm 1.26.11 (latest) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]
    $ trap
    $

tayler1:
Resolves #3540 and #3560

### DIFF
--- a/scripts/functions/environment
+++ b/scripts/functions/environment
@@ -276,6 +276,12 @@ __rvm_teardown()
 
   __variables_definition unset
 
+  if
+    [[ "${_system_name}" == "OSX" ]] && __rvm_version_compare "${_system_version}" -ge 10.11 &&
+    [[ -n "$BASH" ]]
+  then trap 'shell_session_update' EXIT
+  fi
+
   return 0
 }
 


monolar:
This in fact does not properly work for me.

When using iTerm or setting `SHELL_SESSION_HISTORY` to `0` which disables sessions i still constantly get `-bash: shell_session_update: command not found`

I would prefer an actual feature check instead (or additionally since this appears to be still an EL capitan specific thing to some degree) of a platform check. Something like http://stackoverflow.com/a/9529981 in order to check if the function `shell_session_update` is defined.

mcfedr:
I am seeing this bug when I am using gitlab ci, it creates a non login bash session which doesnt have `shell_session_update` so it causes all the builds to fail.

tayler1:
This PL doesn't make sense anymore.
Just upgrade rvm. 
https://github.com/rvm/rvm/commit/cbc4488028ccc682a872034ba823f61bf2d9cf6d

PR #3628, opened 2016-02-18
should deal with ~/.bash_sessions_disable

fanyeren:
shell_session_update not found when ~/.bash_sessions_disable exists

Signed-off-by: feng xiahou xiahoufeng@yahoo.com


mpapis:
well, I would think we should just check `if is_a_function shell_session_update`

fanyeren:
@rys done

PR #3649, opened 2016-03-22
Bump jruby version to 9.0.5.0

### LINKED ISSUE #3648: jruby should default to 9.0.5.0
I have a `.ruby-version` file in my project that looks like this:

```
$ cat .ruby-version
jruby
```

When I cd to this folder, I get the following:

```
$ cd foo
Unknown ruby string (do not know how to handle): jruby-9.0.4.0.
jruby-9.0.4.0 is not installed.
To install do: 'rvm install jruby-9.0.4.0'
```

I do have 9.0.5.0 installed, but it seems to prefer 9.0.4.0 for some random reason. I just did `rvm get head`, but it didn't help. Tried to poke around in this repo but couldn't find any obvious place where `jruby` is pointed at `9.0.4.0`. Any ideas/suggestions? TIA.

rys:
Fixes #3648


perlun:
Thanks! Working perfectly now. :clap:

PR #3653, opened 2016-03-29
avoid infinite loop

Niandalu:
I'm using zsh + tmux

when the working dir is removed, the $working_dir is going to be "."
which causes the infinite loop while calling __rvm_project_dir_check below.
you can reproduce it easily by using tmux.

### How to reproduce:
1. Let's add an line `echo "loop!!"` to `scripts/functions/rvmrc_project:36`
2. create a dir `/tmp/a`
3. **`cd /tmp/a`**, then run `tmux new-session -s test`
4. use the shortcut to `detach-client`
5. `..; rm -rf a`
6. use the shortcut to `split-window`
7. Now you got an infinite loop

loop

### DIFF
--- a/scripts/functions/rvmrc_project
+++ b/scripts/functions/rvmrc_project
@@ -20,7 +20,7 @@ __rvm_project_rvmrc()
   while :
   do
     if
-      [[ -z "$working_dir" || "$HOME" == "$working_dir" || "${rvm_prefix:-}" == "$working_dir" ]]
+      [[ -z "$working_dir" || "$HOME" == "$working_dir" || "${rvm_prefix:-}" == "$working_dir" || "$working_dir" == "." ]]
     then
       if (( ${rvm_project_rvmrc_default:-0} >= 1 ))
       then rvm_previous_environment=default


mpapis:
I was not able to reproduce it from normal shell, before we merge it - maybe this should be fixed elsewhere? maybe `rvm/scripts/rvm`?

Niandalu:
@mpapis Maybe I didn't make it clear enough, so I just added `cd /tmp/a` to step 3. The issue only happens when I'm using tmux. The directory where you create a new tmux session will work as a default working directory for tmux. And when you try to create any new panel/window, it will try to access the working directory. If the directory is not existed anymore, the path of the directory will set to `.` which causes the loop.

And as for the fix suggestion, the reason I put the line into `rvm/scripts/functions/rvmrc_project` is because:
1. it happens only when rvm_project_rvmrc is enabled
2. and the function module(`rvm_project`) should be the only one knows what's the current working directory, where we can check whether the working directory existence

I'm not familiar with Shell, if I'm wrong, please let me know :P Thanks a lot

mpapis:
now it makes sense, thanks for the extra explanation.

PR #3666, opened 2016-04-18
Enable building of Ruby 1.9.3 and 2.0.0 on 64bit little-endian PowerPC

daxtens:
When trying to compile Ruby 1.9.3 and 2.0.0 for ppc64le using rvm, we get the
following errors:

```
Error running './configure --prefix=/home/builder/.rvm/rubies/ruby-1.9.3-p551 --disable-install-doc --enable-shared',
showing last 15 lines of /home/builder/.rvm/log/1460701268_ruby-1.9.3-p551/configure.log
/usr/bin/uname -p =
/bin/uname -X     =

hostinfo               =
/bin/universe          =
/usr/bin/arch -k       =
/bin/arch              =
/usr/bin/oslevel       =
/usr/convex/getsysinfo =

UNAME_MACHINE = ppc64le
UNAME_RELEASE = 4.5.0-00008-ga6de155
UNAME_SYSTEM  = Linux
UNAME_VERSION = #20 SMP Sat Apr 2 22:28:04 AEDT 2016
configure: error: cannot guess build type; you must specify one
There has been an error while running configure. Halting the installation.
```

Looking at the full log, we can see that this is due to the
`config.sub` and `config.guess` files included with old Ruby 
versions being too old to know about ppc64le as a platform.

To fix this, we grab the latest files from gnu.org, and create a
patch that brings the included files up to date.

With this patch, all the 1.9.3 and 2.0.0 variations build.

Upgrading these files unconditionally is safe, as they're only used
to detect the build platform. (Distro package build processes, for
example, tend to upgrade these files automatically.)


daxtens:
Many thanks :)

PR #3667, opened 2016-04-20
introduction of Ruby v2.1.10

ncreuschling:
Smoother use of RVM with an already installed Ruby v2.1.10 when using an .rvmrc file with "rvm use 2.1"


havenwood:
It looks like we'll also need checksums for 2.1.9 and 2.1.10 in `config/md5` and `config/sha512`.

rys:
Fixed by #3672

PR #3670, opened 2016-04-27
switch another url to https

emirikol:
this seems to have been looked over in a previous commit, the previous command results in downloading an html redirection page which isn't very useful. to say nothing of the possible security risk.


reedloden:
Note that this was a purposeful decision in #3469, FYI. See https://github.com/rvm/rvm/pull/3469#discussion_r35320295 for some of the discussion. However, it seems @bagder has since made his site https://, which defeats the whole point of using http:// here in the first place. However, I am of the opinion that it's the OS's job to get the initial bootstrapping of certificate authorities correct, so not sure rvm should support a broken system like that anyway. Anyway, just giving some backstory here. :)

PR #3741, opened 2016-08-30
Adding tests for OpenJDK 1.8.0 for Jruby + CentOS/Fedora

tadman:
This impacts Fedora where Fedora 24 ships with OpenJDK 1.8.0 only, 1.7.0 is no longer available as a package. Jruby 9.1.2.0 is happy with 1.8.0 so printing an advisory about 1.7.0, which is very difficult to install manually, is misleading.

Presumably this is good to get in now since RHEL and CentOS will update to this soon if they haven't already.


pkuczynski:
Thanks!

Issue #1746, opened 2013-03-30
command not found: __rvm_cleanse_variables


pawlyk:
I use MRI 2.0.0-p0 installed through RVM. When i run shell i see  following error:
`/home/pasha/.rvm/scripts/initialize:41: __rvm_cleanse_variables: function definition file not found`
I'm using ZSH (zsh 5.0.0 (x86_64-unknown-linux-gnu))  on Ubuntu 12.10(amd64)

% rvm -v
`rvm 1.19.0 (master) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]`
After installed ruby i added line to my .zshrc:
`[[ -s "$HOME/.rvm/scripts/rvm" ]] && . "$HOME/.rvm/scripts/rvm"`
Sorry if this is a duplicate.

Issue #770, opened 2012-02-20
command not found: __rvm_cleanse_variables (again?)


mysteriouspants:
I see #662 is closed, but I'm getting the same problem with MRI 1.9.3-p125:

```
/Users/cmiller/.rvm/scripts/initialize:40: __rvm_cleanse_variables: function definition file not found
```

I'm using ZSH (`zsh 4.3.11 (i386-apple-darwin11.0)`) on OS 10.7.3 with oh-my-zsh. I tried reverting back to stable, upgrading to head, and reinstalling RVM (by moving the old .rvm to .rvm-backup and then installing again), but it still gives me this error. I'm also experiencing issues where Rails is loosing its head and refusing to run the server (the `rails s` command tells me "no Ruby script found in input (LoadError)"). RVM version says it's at 1.10.2.

```
> rvm -v

rvm 1.10.2 by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.beginrescueend.com/]
```

I'm a little interested in why it's doing this. I think it's plenty possible that I did something wrong. Sorry if this is a duplicate!

dwcaraway:
the ```rm -f ~/.zcompdump*``` command to clear zcompdump worked for me as well, thanks @tam-vo

johnwatsondev:
@tam-vo  Thanks for share.

casatwy:
rm -f ~/.zcompdump* does not work for me.

I checked the ~/.zlogin and modfied the rvm script path to below:

`[[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm" # Load RVM into a shell session *as a function*`

and it works.

I think you should check wether there is rvm in `~/.rvm/scripts`

wesson-yi:
@Jeweller-Tsai， you are very clever。I don't know why, but I can avoid him.(￣.￣)

Issue #662, opened 2011-12-26
command not found: __rvm_cleanse_variables when loaded from .zshrc


vertis:
The latest version of rvm (3f3ffc4d2d30ed151fc308e4ed253c60102878e1) results in the following errors: 

```
/Users/vertis/.rvm/scripts/initialize:40: command not found: __rvm_cleanse_variables
/Users/vertis/.rvm/scripts/rvm:140: command not found: __rvm_setup
/Users/vertis/.rvm/scripts/rvm:152: command not found: __rvm_teardown
```

This happens when it is sourced from .zshrc

PR #3753, opened 2016-09-12
Better Zsh compinit compatibility

zeorin:
@mpapis

In Zsh, compinit creates a `~/.zcompdump` file to cache completions. The
reading of this dump file by the shell creates all the functions in it
as *undefined functions*, which are functions that are actually defined
to be autoloaded when executed (sounds weird, I know). That autoloading
only works if that function can be found in Zsh's `fpath`.

At the first run of RVM in a shell session, `__rvm_cleanse_variables`
will not yet have been defined nor be in an `fpath` somewhere, **but**
it will have been defined as an undefined function if it's in the
`~/.zcompdump` file.

Most users of Zsh will simply run `compinit` somewhere in their
`.zshrc`. Invoking it in that way will *recreate* the `~/.zcompdump`
file from scratch, which means that in that case
`__rvm_cleanse_variables` will not yet have been defined and it's
possible to determine whether to attempt to run it by checking if it's a
function.

*However*, some users of Zsh do not run `compinit` that way. Some run it
manually, others will only run it once a day in a way that recreates the
`~/.zcompdump` file from scratch, simply reusing an existing one for the
rest of the day (using `compinit -C`, see here:
http://carlosbecker.com/posts/speeding-up-zsh/).

This commit simply attempts to run the function whether it has been
defined or not, redirecting `stderr` to `/dev/null`. This way RVM can
play nice with various `compinit` strategies and issues like #1746, #770
and #662 shouldn't happen again.

There might be a slight issue with redirecting `stderr` to `/dev/null`,
as presumably this will redirect all error output (even errors from
functions that `__rvm_cleanse_variables` calls) to `stderr`. If that is
undesirable, perhaps there's a way to determine whether
`__rvm_cleanse_variables` has been defined as an undefined function, or
perhaps whether it's really the first time `initialize` is being run
instead of as a result of `rvm reload` or something.

### DIFF
--- a/scripts/initialize
+++ b/scripts/initialize
@@ -42,14 +42,9 @@ export __rvm_env_loaded
 if [[ -z "${rvm_tmp_path:-}" ]] || (( __rvm_env_loaded == 1 ))
 then
 
-  if
-    \typeset -f __rvm_cleanse_variables >/dev/null 2>&1
-  then
-    __rvm_cleanse_variables ||
-    if [[ -n "${ZSH_VERSION:-}" ]]
-    then rvm_error "ZSH Error, run \`rm -f ~/.zcompdump\` and try again."
-    fi
-  fi
+  # Because of compinit in ZSH, this function may not be defined yet
+  # even though it may exist.
+  __rvm_cleanse_variables 2>/dev/null
 
   # Ensure that /etc/rvmrc and $HOME/.rvmrc values take precedence.
   if (( ${rvm_ignore_rvmrc:=0} == 0 ))


mpapis:
could you move `__rvm_cleanse_variables` code to this place, I do not like suppressing the output, the function is used in just here, so we could as well have it in place :)

zeorin:
For the life of me, I cannot reproduce this issue after doing a `rvm get stable`. I should really have done that before submitting this pull request. My fix is no longer necessary, `__rvm_cleanse_variables` is not added to `.zcompdump` anymore.

PR #3767, opened 2016-09-21
Use Standard `command` instead of `which`

kojiromike:
If a system is able to run these scripts, it is guaranteed to have the `command` builtin. It is not guaranteed to have the `which` command. If it does, the `which` command is not guaranteed to behave consistently.

Despite Herculean effort to avoid shell aliases from altering the behavior, shell functions can still have a significant impact. The below is what brought this to my attention:

```
Downloading https://github.com/rvm/rvm/archive/1.27.0.tar.gz
Downloading https://github.com/rvm/rvm/releases/download/1.27.0/1.27.0.tar.gz.asc
rvm-installer: line 404: gpg2 is /usr/local/bin/gpg2: No such file or directory
Warning, RVM 1.26.0 introduces signed releases and automated check of signatures when GPG software found.
```

because

```
$ which which
which is a function
which ()
{
    type -a $1
}
which is /usr/bin/which
```

### DIFF
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -30,8 +30,7 @@ fail() { log "\nERROR: $*\n" ; exit 1 ; }
 
 rvm_install_commands_setup()
 {
-  \which which >/dev/null 2>&1 || fail "Could not find 'which' command, make sure it's available first before continuing installation."
-  \which grep >/dev/null 2>&1 || fail "Could not find 'grep' command, make sure it's available first before continuing installation."
+  __rvm_which grep > /dev/null || fail "Could not find 'grep' command, make sure it's available first before continuing installation."
   if
     [[ -z "${rvm_tar_command:-}" ]] && builtin command -v gtar >/dev/null
   then
@@ -59,9 +58,9 @@ rvm_install_commands_setup()
             log "Trying to install GNU version of tar, might require sudo password"
             if (( UID ))
             then
-              if \which sudo >/dev/null 2>&1
+              if __rvm_which sudo >/dev/null
               then sudo_10=sudo
-              elif \which /opt/csw/bin/sudo >/dev/null 2>&1
+              elif __rvm_which /opt/csw/bin/sudo >/dev/null
               then sudo_10=/opt/csw/bin/sudo
               else fail "sudo is required but not found. You may install sudo from OpenCSW repository (https://www.opencsw.org/about)"
               fi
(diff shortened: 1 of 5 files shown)


pkuczynski:
Your suggestion sounds reasonable. Can you please first have a look at the failing build and then I will check in details if we can merge it?

kojiromike:
@pkuczynski happy to. I looked a the build logs and they don't seem to be related to the changes I made. Two of them look like Travis timed out. Is there some way to re-trigger the tests?

pkuczynski:
I just restarted the build. Results here: https://travis-ci.org/rvm/rvm/builds/161625909

binscripts/rvm-installer
```diff
@@ -219,7 +218,7 @@ __rvm_curl()
 )
 
 rvm_error()  { printf "ERROR: %b\n" "$*"; }
-__rvm_which(){   which "$@" || return $?; true; }
+__rvm_which(){ command -v "$@" 2>/dev/null; }
```
mpapis: please change to:
```bash
__rvm_which(){ builtin command -v "$@" 2>/dev/null || return $?; true; }
```
this will prevent tinkering with the command, and will fix problems with shell flags enforcing script fail on anyt command fail

also please note that `__rvm_which` is also used as the `which` to find files on `PATH` not functions, for example:
```bash
scripts/functions/osx-ssl-certs:95:  ssl_binary="${1:-$(__rvm_which openssl)}"
scripts/functions/env:140:  while local_rvm_path="$( __rvm_which rvm 2>/dev/null )"
```
this functionality has to be preserved, unfortunately neither `command -v` or `type -a` show you the path to a program if it's hidden with a function / alias

any ideas?

kojiromike: If we're guaranteed bash (not just sh), we can use `type -P` to get the path to an external command. As for `builtin`, it's just as easy to goof around with that as `command`, right?

```builtin() { eval "$1() { echo lol; }"; }```

havenwood: `type -P` isn't portable to zsh: `zsh: bad option: -P`

kojiromike: @havenwood All of the files in this PR have the shebang `#!/usr/bin/env bash`, so zsh compatibility should not be a requirement. If it is, then the shebang should probably read `!#/bin/sh`.

mpapis: but we want to keep the same code in the other places, so `__rvm_which` function in the other place - would need to be changed, and then we get a problem, as there we can work with `zsh`

Issue #3785, opened 2016-10-24
Can't download rvm; curl returned status '23'


Mersion:
```
C:\Users\Me> bash rvm-installer stable
Downloading https://github.com/rvm/rvm/archive.1.27.0.tar.gz
curl: (23) Failed writing body (0 != 762)

Could not download 'https://github.com/rvm/rvm/archive.1.27.0.tar.gz'
curl returned status '23'

Downloading https://bitbucket.org/mpapis/rvm/get/1.27.0.tar.gz
curl: (23) Failed writing body (0 != 13992)

Could not download 'https://bitbucket.org/mpapis/rvm/get/1.27.0.tar.gz'
curl returned status '23'
```
I'm trying to install rvm but i can't. I've tried different solutions but without success. Please help me find a solution.

pkuczynski:
I can see you are using Windows? Why don't you install RVM on cygwin?

Mersion:
On cygwin i get:
`$ bash rvm-installer stable 
bash: rvm-installer: No such file or directory`

pkuczynski:
Can you please run `curl -L https://get.rvm.io | bash -s master`

Mersion:
![bez nazwy](https://cloud.githubusercontent.com/assets/23041908/19689724/997c06ec-9ace-11e6-9afa-628af1d26171.png)

pkuczynski:
Please gist output of `curl -L https://get.rvm.io | bash -s master --trace`

pkuczynski:
It looks like there is permission problem...

Mersion:
I have send link to you on email.

Mersion:
gist.github.com gist.github.com  
                
              
               
                 
                  Dnia 25 października 2016 16:33 Piotr Kuczynski &lt;notifications@github.com&gt; napisał(a):
                 
                 
                   It looks like there is permission problem...

Mersion:
https://gist.github.com/Mersion/f11555e729e15c5451812a0c34c777e3/revisions

pkuczynski:
Can you try `curl -LN https://get.rvm.io | bash -s master --trace`

Mersion:
Sorry, nothing changed. ;<

pkuczynski:
Can you run in your console:

[...]

PR #3794, opened 2016-11-01
Attempt to fix curl returning status `23` on Windows #3785

pkuczynski:


sikachu:
This PR made signature of RVM installer invalid. @mpapis do you mind re-sign rvm-installer and uploads it to the repo?

PR #3826, opened 2016-11-27
Remove slash from curl command of install section

markelog:
It doesn't seem that it's needed?


mpapis:
the backslash is needed to prevent aliased versions, usually this can be used to add extra options which can disrupt the process

markelog:
Oh, okay, cool, sorry for the noise then

PR #3836, opened 2016-12-07
on osx update brew and install zlib

cokochiaki:


pkuczynski:
I had a branch stale for a while which was already adding zlib on the RVM level: https://github.com/rvm/rvm/pull/3837

So I would reject this PR. Thank you for your effort and reminding me about the old branch.

PR #3841, opened 2016-12-15
Bump MRI default to 2.3.3

listrophy:
I'm not sure that this is everything that needs to be done, but it's my best guess.


pkuczynski:
Your code changes are correct, but created against wrong branch. Development happens in master and we already did all of your suggested changes. Its only waiting for release...

PR #3849, opened 2016-12-25
add ruby 2.4.0

vincentaubert:


pkuczynski:
Missing 'config/db' update and entry in CHANGELOG

ur5us:
@pkuczynski I don't think anything needs to be added to `config/db` but an entry to the CHANGELOG is missing indeed.

pkuczynski:
There is a line listing latest MRI ruby. It now points to 2.3.3

ur5us:
@pkuczynski Ah, I see, thanks for clarifying it.

vincentaubert:
fixed

PR #3855, opened 2016-12-30
Allow to check for undesired packages and remove them

pkuczynski:
Fixes #3851

PR #3857, opened 2016-12-31
Remove incompatible version of openssl098

### LINKED ISSUE #3844: Remove incompatible version of openssl098
Run `brew unlink openssl098` when old version of openssl detected

See #3656 for more context

pkuczynski:
Fixes #3844
Uses #3855

### DIFF
--- a/scripts/functions/requirements/osx_brew
+++ b/scripts/functions/requirements/osx_brew
@@ -49,17 +49,25 @@ requirements_osx_brew_lib_available()
   brew search | __rvm_grep "^${1}$" >/dev/null || return $?
 }
 
+requirements_osx_brew_libs_error()
+{
+    rvm_warn "There were package ${1} errors, make sure to read the log.
+
+Try \`brew tap --repair\` and make sure \`brew doctor\` looks reasonable.
+
+Check Homebrew requirements https://github.com/Homebrew/homebrew/wiki/Installation"
+}
+
+}
 requirements_osx_brew_libs_install()
 {
   brew unlink  "$@" || true # ignore: Error: No such keg: /usr/local/Cellar/automake
   brew install "$@" ||
   {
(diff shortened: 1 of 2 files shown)


scripts/functions/requirements/osx_brew
```diff
@@ -49,17 +49,25 @@ requirements_osx_brew_lib_available()
...
+requirements_osx_brew_libs_error()
+{
+    rvm_warn "There were package ${1} errors, make sure to read the log.
+
+Try \`brew tap --repair\` and make sure \`brew doctor\` looks reasonable.
+
+Check Homebrew requirements https://github.com/Homebrew/homebrew/wiki/Installation"
+}
+
+}
```
mpapis: double closing of the function?

scripts/functions/requirements/osx_brew
```diff
@@ -290,6 +311,8 @@ requirements_osx_brew_libs_default()
 
   requirements_osx_brew_check_custom homebrew/dupes # for: zlib
 
+  undesired_check openssl098
```
mpapis: not sure it should be in `custom` or rather in `before` or `define`? or maybe we should add `before_install`? or even a new step `uninstall`? - if `custom` is called always - just leave it :)

scripts/functions/requirements/osx_brew
pkuczynski: good catch! :)

scripts/functions/requirements/osx_brew
pkuczynski: Yeah, the `requirements_osx_brew_libs_default` is called in few places, that's why I put it there. Not sure if we should have a dedicated `uninstall` function. Possibly. Didn't want to invest too much of the time having RVM2 down the road...

PR #3859, opened 2016-12-31
Add comment about rvm PATH being the las

### LINKED ISSUE #3854: Add comment about rvm PATH being the last
In .bash_profile and similar shell config files, it would be great if rvm adds a comment notice 
that this line should come last after all other PATH variable changes.
Additionally, the relevant line should be automatically moved to the last line or 
offered as an option if there are issues with PATH order.

pkuczynski:
Fixes #3854

### DIFF
--- a/scripts/functions/installer
+++ b/scripts/functions/installer
@@ -877,7 +877,7 @@ then
     source \"/usr/local/rvm/scripts/rvm\"
   fi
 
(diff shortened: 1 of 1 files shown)


mpapis:
@strarsis What do you think?

strarsis:
@mpapis: Yes, the added comment helps e.g. when adding new PATH lines manually.

PR #3860, opened 2016-12-31
Missing libyaml-devel on PCLinuxOS 64-bit

### LINKED ISSUE #3703: Missing libyaml-devel on PCLinuxOS 64-bit
Tried to install ruby-2.3.1 on PCLOS 64-bit and got this error:

```
rvm_error "There is no 'libyaml-devel' package available for installation and RVM does not know how to make that happen, please tell us how to install 'libyaml-devel' on 'PCLinuxOS-2016' here: https://github.com/rvm/rvm/issues
```

For PCLOS 64-bit the library is named:  "lib64yaml-devel"
This is true for all the 64-bit libraries which have 32-bit versions.

pkuczynski:
Fixes #3703

### DIFF
--- a/scripts/functions/requirements/pclinuxos
+++ b/scripts/functions/requirements/pclinuxos
@@ -3,17 +3,12 @@
 requirements_pclinuxos_lib_installed()
 {
   rpm -q --whatprovides "$1" >/dev/null || return $?
+  # Alternative: rpm -qa | grep "$1" >/dev/null || return $?
 }
 
 requirements_pclinuxos_lib_available()
 {
-  synaptic list --cacheonly --quiet "$1" >/dev/null 2>&1 || return $?
-}
-
-# required only for libyaml-devel check as fedora is broken ...
-requirements_pclinuxos_lib_available_no_caching()
-{
-  synaptic list --quiet "$1" >/dev/null 2>&1 || return $?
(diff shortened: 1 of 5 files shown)


scripts/functions/requirements/pclinuxos
```diff
@@ -45,20 +57,21 @@ or there was problem checking if libyaml-devel is available / installed."
 # not sure if this one is needed for pclinuxos, copied from centos
 requirements_pclinuxos_update_system_check_epel()
```
pkuczynski: I actually think this function is not implemented in the right way. Issue reports `PCLinuxOS-2016` and we do not have `epel2016` in `config/db`. So we should figure out another way of checking if system needs to be uodated (or remove the `[[ -z "$epel_rpm" || -z "$epel_key" ]]` part)

mpapis: or we can add `epel2016` - ask people from the tickets that triger that change what would be the best values to use :)

pkuczynski: epel2016 does not exist. The latest is epel7, which I just added, but it's not gonna solve our issue.

mpapis: one of them has to be compatible? maybe the older or newer one?

PR #3870, opened 2017-01-08
Fix typo

hashedhyphen:
RMV -> RVM


hashedhyphen:
I'm sorry, the account of this commit is wrong... I will amend it.

PR #3881, opened 2017-01-17
Remove kernel-libc-devel dependency on Solus

theSoenke:
Small fix for Solus. The package kernel-libc-devel package has been renamed to kernel-headers and causes rvm install to fail because kernel-libc-devel cannot be found by the package manager. The package is also not required and can be removed

* Remove kernel-libc-devel dependency on Solus


pkuczynski:
And where is the CHANGELOG entry?

pkuczynski:
Never mind, I just added it... :(

theSoenke:
Thanks for adding the changelog. Missed it somehow

PR #3892, opened 2017-01-25
Print autolibs flag value in `rvm info`

pkuczynski:


scripts/functions/build_config
```diff
@@ -19,12 +19,12 @@ __rvm_setup_compile_environment()
 
 __rvm_setup_compile_environment_setup()
 {
-  if [[ -z "${rvm_autolibs_flag:-}" ]]
```
mpapis: we need this `if` - user could have `export rvm_autolibs_flag=bla` in their shell configuratio

PR #3908, opened 2017-02-07
Detect `noexec` mount mode for partition hosting RVM home

### LINKED ISSUE #3832: Detect `noexec` mount mode for partition hosting RVM home
## Description

Installing in user mode leads to a broken install with message "permission denied" on almost all commands.

## Steps to reproduce

1. I'm installing using a dedicated user (and as a single-user install), with:
   `\curl -sSL https://get.rvm.io | bash -s -- --autolibs=read-fail`.
2. Then I'm disconnecting & reconnecting.

## Expected behavior

`rvm help` should show the help.

## Actual behavior

`rvm help` gives: `-bash: /home/myuser/.rvm/scripts/help: Permission denied`.

The same error occurs with most command (e.g. `rvm list`). `rvm version` works, though.

I noticed that most scripts under `/home/myuser/.rvm/scripts` do not have the execution bit set (unlike on another deployment via Vagrant).

## Environment info

Ubuntu Xenial (but I've also seen another deployment where this issue doesn't occur).

```
rvm info
-bash: /home/myuser/.rvm/scripts/info: Permission denied
```

pkuczynski:
Fixes #3832


binscripts/rvm-installer
```diff
@@ -781,6 +781,24 @@ follow this link for details how to fix:
   fi
 }
 
+rvm_install_validate_volume_mount_mode()
+{
+  \typeset partition
+  \typeset mount_noexec_mode
+
+  partition=`df -P $rvm_path | tail -1 | cut -d' ' -f 1`
+  mount_noexec_mode=`mount | grep -q "$partition.*noexec.*"`
```
mpapis: `mount | grep -q "$partition .*noexec.*"`` - extra space to prevent matching wrong mount point

pkuczynski: Fixed

PR #3922, opened 2017-02-10
Adds binaries for Ubuntu 16.10

### LINKED ISSUE #3823: Ubuntu 16.10 binaries
Add the following binaries from https://rvm.io/binaries/ubuntu/16.04/x86_64/ but for Ubuntu 16.10

or, is there any alternate way to install ruby via rvm on Ubuntu 16.10?

jrgifford:
Fixes #3823 

Adds a whole slew of Ruby binaries for Ubuntu 16.10 (which are [uploaded to rvm.io](https://rvm.io/binaries/ubuntu/16.10/x86_64/)).


mpapis:
this is not needed anymore?

pkuczynski:
@jrgifford I think you unnecessarily sorted the files, which makes it very hard to review the changes. Could you please update your PR to include only necessary updates?

jrgifford:
All I did was run the binary checksum script, which shouldn't cause any issues.


That said, I don't think this PR is required any more, but I'll check a little later today and get back to y'all.

pkuczynski:
@jrgifford which script exactly did you run?

jrgifford:
@pkuczynski I ran this one: https://github.com/rvm/rvm/blob/master/update-remote.sh

jrgifford:
I ran that script, and then manually added the changes to the changelog.

jrgifford:
@pkuczynski Any thoughts? I think the script just makes really noisy diffs.

pkuczynski:
I cleaned up this pull request and I wanted to dbl check - why do you add osx 10.12 files with this PR? Why those files on the server are so old? @mpapis any ideas?

jrgifford:
I don’t now RE: why those were added. Most likely the checksums weren’t added when the binaries were built.

PR #3933, opened 2017-02-20
Use actual executable test instead of mount|grep noexec

### LINKED ISSUE #3832: Detect `noexec` mount mode for partition hosting RVM home

gnosek:
Fixes #3832.

* Don't try to parse mounted filesystem table to detect partitions mounted `noexec`
* Create a temporary script and try to execute it to actually test if scripts are runnable from the target location

### DIFF
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -783,7 +783,7 @@ follow this link for details how to fix:
 
 rvm_install_validate_volume_mount_mode()
 {
-  \typeset path partition
+  \typeset path partition test_exec
 
   path=$rvm_path
 
@@ -793,10 +793,14 @@ rvm_install_validate_volume_mount_mode()
   do
       if [[ -d $path ]]
       then
+        test_exec=$(mktemp $path/rvm-exec-test.XXXXXX)
+        echo '#!/bin/sh' > "$test_exec"
+        chmod +x "$test_exec"
         partition=`df -P $path | awk 'END{print $1}'`
 
-        if mount | grep "$partition .*noexec.*" >/dev/null 2>&1
+        if ! "$test_exec"
         then
+          rm -f "$test_exec"
           printf "%b" "
 It looks that partition holding RVM destination location ${rvm_path}
 is mounted in *noexec* mode, which prevents RVM from working correctly.
@@ -804,6 +808,7 @@ Please re-mount partition ${partition} without the noexec option."
           exit 2
         fi
 
+        rm -f "$test_exec"
         break
       fi
 
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -787,24 +787,25 @@ rvm_install_validate_volume_mount_mode()
 
   path=$rvm_path
 
-  # $rvm_path might not exists at this point so we need to traverse the directory tree upwards
-  # to find the name of the partition and then check if it has not been mount with noexec mode
+  # Directory $rvm_path might not exists at this point so we need to traverse the tree upwards
   while [[ -n "$path" ]]
   do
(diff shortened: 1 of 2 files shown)


binscripts/rvm-installer
```diff
@@ -793,17 +793,22 @@ rvm_install_validate_volume_mount_mode()
...
         partition=`df -P $path | awk 'END{print $1}'`
 
-        if mount | grep "$partition .*noexec.*" >/dev/null 2>&1
+        if ! "$test_exec"
         then
+          rm -f "$test_exec"
           printf "%b" "
 It looks that partition holding RVM destination location ${rvm_path}
 is mounted in *noexec* mode, which prevents RVM from working correctly.
 Please re-mount partition ${partition} without the noexec option."
```
mpapis: we could rephrase it with `Please re-mount ${path} without the noexec option.` - and drop the partition detection - one less thing to think about.

binscripts/rvm-installer
```diff
@@ -783,27 +783,33 @@ follow this link for details how to fix:
...
 {
-  \typeset path partition
+  \typeset path partition test_exec
 
   path=$rvm_path
 
-  # $rvm_path might not exists at this point so we need to traverse the directory tree upwards
-  # to find the name of the partition and then check if it has not been mount with noexec mode
+  # Directory $rvm_path might not exists at this point so we need to traverse the tree upwards
   while [[ -n "$path" ]]
```
pkuczynski: I am wondering if we should traverse the tree or rather create the folder? @mpapis what you think?

gnosek: Yeah, I missed the comment, sorry for that. Me, personally, I'd create a script in the final location (~/.rvm/bin or similar, creating directories as required) but I didn't want to mess with the internals too much (maybe something later checks if the directory exists and would get confused).

Then we could also use a hardcoded script name and not rely on `mktemp` (which should be pretty universal but given you check for `grep` and `which` I understand you want minimal external dependencies)

binscripts/rvm-installer
gnosek: +1 from me, FWIW. The partition can be mounted in several places at once with different flags (which is what actually caused issues for me in the first place).

I can fix that but call me stupid but I don't know how to update the pull request without dropping changes by @pkuczynski

binscripts/rvm-installer
mpapis: I was also thinking about `mkdir -p` but this would check if we can create the path also - which would require more error handling and could leave garbage directories if user decides not to use this path... I'm fine with traversing the path down, in the end write permissions are required to create the directory so the script should be able to also create the script, it's easier to cleanup one file.

pkuczynski:
@mpapis could you sign the installer before we merge it in?

PR #3961, opened 2017-03-22
Detect `.` as an alternative to `source` in bash profile warning

### LINKED ISSUE #3960: Spurious warning that `.profile` is not being sourced when using the `.` source alias
## Description

I was getting warned that I wasn't sourcing `~/.profile` from my `.bash_profile` because I was using the `.` alias for `source` and the regular expression in `scripts/notes` was not created to detect this.

## Steps to reproduce

This occurred while updating rvm itself. I have a `~/.profile` file and my `.bash_profile` is set up to source it with:

    . $HOME/.profile

 1. run `rvm get stable`
 2. wait for it to complete

## Expected behavior

I did not expect to be warned to source `.profile` from `.bash_profile` because I am already doing this.

## Actual behavior

I get the following warning:

> * WARNING: You have '~/.profile' file, you might want to load it,
>    to do that add the following line to '/Users/cowen/.bash_profile':
>
>      source ~/.profile

oysta:
Add a check for the `.`  source syntax in the bash startup files before warning about `.profile` not being sourced. 

Fixes #3960

* Use a regular expression that also detects the use of `.` as well as `source`
* Improved the wording of the warning a little

### DIFF
--- a/scripts/notes
+++ b/scripts/notes
@@ -204,8 +204,7 @@ $notes_type Notes:
         __rvm_grep -E '(source|\.).+profile' "${__var}" >/dev/null 2>&1 ||
         {
           important_message "
-  * WARNING: You have a '~/.profile' file, you might want to load it.
-    To do that add the following line to '${__var}':
+  * WARNING: '~/.profile' file found. To load it into your shell, add the following line to '${__var}':
 
       source ~/.profile
 "
--- a/scripts/notes
+++ b/scripts/notes
@@ -201,7 +201,7 @@ $notes_type Notes:
       if
         [[ -f "${__var}" ]]
       then
-        __rvm_grep 'source.*profile' "${__var}" >/dev/null 2>&1 ||
+        __rvm_grep -E '(source|\.).+profile' "${__var}" >/dev/null 2>&1 ||
         {
           important_message "
   * WARNING: You have '~/.profile' file, you might want to load it,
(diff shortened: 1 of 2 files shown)


pkuczynski:
Thank you for your PR! It's always welcome to see community contributions.

The problem with this PR is that it solves your very specific case, where you aliased `source` with `.`. This is not something common. The correct way of fixing it would be to grab the first group and check if it `== source` or if it is an alias of `source`.

oysta:
Sorry, calling it an "alias" is a bit ambiguous. The `.` is built in `sh` syntax, not something I've created myself. (In fact I believe `.` was the original way to source scripts back in the days of old)

http://askubuntu.com/questions/25488/what-is-the-difference-between-source-and-in-bash

pkuczynski:
Ah, I see. Didn't know. Then I think its OK :)

scripts/notes
```diff
@@ -201,11 +201,11 @@ $notes_type Notes:
       if
         [[ -f "${__var}" ]]
       then
-        __rvm_grep 'source.*profile' "${__var}" >/dev/null 2>&1 ||
+        __rvm_grep -E '(source|\.).+profile' "${__var}" >/dev/null 2>&1 ||
         {
           important_message "
-  * WARNING: You have '~/.profile' file, you might want to load it,
-    to do that add the following line to '${__var}':
+  * WARNING: You have a '~/.profile' file, you might want to load it.
```
mpapis: it's not `a profile file`, it's `the '~/.profile' file` - at least I think so, I'm new to the distinction ;)

pkuczynski: better?

oysta:
Thanks everyone 🎉 rvm is a great tool—keep up all the good work!

PR #3969, opened 2017-03-27
Allow users to specify irb history file using IRB.conf[:HISTORY_FILE]

### LINKED ISSUE #3968: Unable to save project-specific irb history
## Description

I am not able to use project-specific irb history files because rvm hard codes the history file path and name.

## Steps to reproduce

1. create `~/.irbrc`
``` # ~/.irbrc
IRB.conf[:HISTORY_FILE] = File.expand_path(Dir.pwd, '.irb-history')
```

2.
```
$ cd ~
$ cd my-project
$ irb
> puts 'hello'
> exit
```

## Expected behavior

I expect to see `puts 'hello'` and `exit` as the last 2 entries in `~/my-project/.irb-history`.

## Actual behavior

I see `puts 'hello'` and `exit` as the last 2 entries in `~/.irb-history`.

## Environment info

Irrelevant. The offending code is in `scripts/irbrc.rb` on master. I looked right before creating this issue.

## Proposed solution

The easiest solution is to modify the readline setup to use IRB.conf[:HISTORY_FILE] if it exists (and maybe set it if it does not). This would also require that the user's `irbrc` file be loaded before readline is set up. It seems like the readline setup could happen last, with the rvm-specific settings still being loaded prior to the user's irbrc. Basically, the setup order would change to: rvm configuration (so it stays ahead of user's irbrc), load user's irbrc, then configure readline.

ToadJamb:
Fixes #3968

* modify the order of operations in `scripts/irbrc.rb` to facilitate the ability of users to set the irb history file using `IRB.conf[:HISTORY_FILE]`.
* use aforementioned `IRB.conf[:HISTORY_FILE]` setting when doing readline configuration if it has been specified - fall back to current behavior if it has not.


scripts/irbrc.rb
```diff
@@ -73,3 +38,40 @@
...
+  # IRB.conf[:HISTORY_FILE] = File.join(ENV["HOME"], ".irb-history")
+
+  # Use an alternate way to on history saving until save-history is fixed.
+  #
+  #   bug:   http://redmine.ruby-lang.org/issues/show/1556
+  #   patch: http://pastie.org/513500
+  #
+  # This technique was adopted from /etc/irbrc on OS X.
+  histfile = File.expand_path(".irb-history", ENV["HOME"])
+  histfile = IRB.conf[:HISTORY_FILE] || histfile
```
pkuczynski: Would it be better to do it as a singleline:

```
histfile = IRB.conf[:HISTORY_FILE] || File.expand_path(".irb-history", ENV["HOME"])
```

or:

```
histfile = IRB.conf[:HISTORY_FILE]
histfile = File.expand_path(".irb-history", ENV["HOME"]) unless histfile
```

Feels more clear to read :)

pkuczynski:
@ToadJamb thanks for your PR. It's highly appreciated.

I made some comments. @mpapis what you think?

scripts/irbrc.rb
mpapis: it all looks like a trick, please the first lines, would it be enough to use only this change:
```ruby
require "irb/ext/save-history" # not even sure this part is needed.
IRB.conf[:HISTORY_FILE] = File.join(ENV["HOME"], ".irb-history")
```

pkuczynski: @mpapis wasn't this code there before and now only moved inside the file?

ToadJamb: There was no reference to `IRB.conf[:HISTORY_FILE]` in the original file. `IRB.conf[:HISTORY_FILE] = File.join(ENV["HOME"], ".irb-history")` would perpetuate the problem I'm attempting to fix. I think you are correct that `require "irb/ext/save-history"` is no longer required (for 1.9+). It MIGHT be necessary for 1.8?...

I am attempting to solve a very specific problem and was hoping to mitigate introducing other issues by minimizing the changes to this file. It hasn't been touched in 3 years, which would seem to indicate that while it may be simplified, there is nothing 'wrong' with most of what it's doing.

The lines regarding setting the history file that I changed:
Originally:
```
histfile = File.expand_path(".irb-history", ENV["HOME"])
```
to:
```
histfile = File.expand_path(".irb-history", ENV["HOME"])
histfile = IRB.conf[:HISTORY_FILE] || histfile
```
There are 100 ways to write this logic such that it still works. This is admittedly 'hack-ish', but I thought it better illustrated the change that I was making than other options.

My personal preference would really be something like this:
```
default_histfile = File.expand_path(".irb-history", ENV["HOME"])
histfile = IRB.conf[:HISTORY_FILE] || default_histfile
```
But technically, that's performing potentially unnecessary calculations, so if you care about that, something like this would better for performance:
(but 'performance' here is nano-seconds difference, especially since 99% of users will be evaluating the `File.expand_path` expression anyway)
```
histfile = IRB.conf[:HISTORY_FILE] || File.expand_path(".irb-history", ENV["HOME"])
```

mpapis: I'm saying the [whole thing](https://github.com/ToadJamb/rvm/blob/0dfc1c19cf33d84280084a14cd3445598aa4b414/scripts/irbrc.rb#L43-L77) can be simplified to:
```ruby
IRB.conf[:HISTORY_FILE] ||= File.expand_path(".irb-history", ENV["HOME"])
```

pkuczynski: @mpapis do you mean that we do not need to push to `Readline::HISTORY` and dump it to the histfile anymore?

mpapis: indeed, it looks that it was a hack to a problem that nobody even knows what it was

pkuczynski: Based on http://stackoverflow.com/questions/2065923/irb-history-not-working It seems that there was an error in Ruby 1.8. As of now it should be enough to use the line which @mpapis proposed. 

@mpapis why haven't we patched 1.8 directly and instead we fix it in irbrc.rb?

mpapis: @pkuczynski I got no insides into the reasons, but as 1.8 usage is far bellow `5%` we can just clean it up and move on.x

pkuczynski: @ToadJamb what you think? Would you be keen to clean it up and test?

ToadJamb: Sure. I'm not convinced that the readline code can be eliminated, but I'm happy to poke at it and let you know what I find. Obviously, it would be great if that whole chunk of code can be removed.

pkuczynski: Cool! Looking forward to get your feedback :)

ToadJamb: For the record, I'm working on an ubuntu box. I need to set `IRB.conf[:HISTORY_FILE]` and `IRB.conf[:SAVE_HISTORY]` to get irb history to save (meaning all the other readline/irb history code is removed). For me, readline works without `require readline`, as well, which means on my machine, I can boil the readline/irb history code down to 2 lines!

One minor note, though... There was code that eliminated duplicate lines from being saved in the history. The 'default' behavior appears to be that every line gets saved even if it's a duplicate, but I don't think that's a big enough reason to keep all that code.

ToadJamb: I do need the `require "irb/completion"` line, as well, for tab completion.

Also... technically, the `.irbrc.rb` file would only need `IRB.conf[:SAVE_HISTORY]` to be set, as irb will save to a default file of `~/.irb_history` if save history is set.

ToadJamb:
I added a commit to reflect the changes mentioned in the code comments.

mpapis:
looks great, please rebase and squash commits :)

pkuczynski:
We can do rebase and squasing on our side straight from Github! Thanks @ToadJamb!

Issue #4008, opened 2017-05-02
rb_thread_fd_close patch for ruby 2.3.4


masterkain:
https://www.ruby-lang.org/en/news/2017/03/30/ruby-2-3-4-released/

ruby 2.3.4 has a known problem about rb_thread_fd_close.

some mainstream (ubuntu) package maintaners are applying this patch in their packages, should we do this as well?

mpapis:
yes we should, could you open a PR with the patch? let us know if you need any guidance.

PR #4010, opened 2017-05-04
rb_thread_fd_close patch for ruby 2.3.4

masterkain:
attempts to fix https://github.com/rvm/rvm/issues/4008


patchsets/ruby/2.3.4/default
```diff
@@ -0,0 +1 @@
+fix_rb_thread_fd_close
```
mpapis: new line please, shell can get confused without those ;)

PR #4027, opened 2017-05-12
Simplify jruby hashes by removing full URL

pkuczynski:


pkuczynski:
@mpapis some debugging info added to verify this change makes no harm...

PR #4044, opened 2017-05-19
Add a method to output multi-colored messages

pkuczynski:
* Add definition for bright colors and two configuration variables: rvm_code_color, rvm_comment_color
* Colored output for `__rvm_cli_version_check`


mpapis:
So the problem is:
rvm-color
We are running it in `$()` which changes the output checked with `-t` not to be terminal.

I have opened a branch with an change <https://github.com/rvm/rvm/compare/feature/colored-output...feature/colored_output_with_replace?expand=1> it allows <https://github.com/rvm/rvm/blob/b9b1b5d04f834b7e179147fe19fad2a27fb288a0/scripts/functions/cli#L167-L174>:
```bash
rvm_color stderr "RVM version <notify>${disk_version}<reset> is installed, yet version <error>${rvm_version}<reset> is loaded."
```

pkuczynski:
I like `rvm_color` and will merge your branch to my branch. But maybe we should call it `rvm_color_out`?

I have a feeling that we have too many functions to output the information to the user:
* rvm_printf_to_stderr - check if user wants colors and if output goes to terminal
* rvm_error - display error
* rvm_warn - display warning
* rvm_log - display info (unless rvm_quiet_flag==1)
* rvm_out - just printf
* rvm_output - printf to selected stream
* rvm_color - printf in color

Don't you think it's a but too much?

mpapis:
I think most of the functions for output could be merged to just 3 as you pointed out and then be just wrappers for `rvm_color_out` example:
```ruby
rvm_error()
{
  rvm_color_output stderr "<$rvm_error_color>$*"
}
```
or something like this :) It might require a loop in the replacement of colors function.

pkuczynski:
@mpapis I merged your changes and made some additional improvements. However for whatever reason sometimes `rvm_warn` for example provides b/w output. Like when I locally run `./install && rvm reload` it shows the "Warning! PATH is not properly set up..." striping all color directives, but not showing any color. Any idea what could be the reason for this behavior?

mpapis:
`__rvm_path_match_gem_home_check` is done before `__rvm_initialize` and the colors are not yet set

scripts/functions/logging
```diff
@@ -87,69 +104,121 @@ rvm_pretty_print()
...
 {
   printf "$@" >&6
 }
+
 rvm_error()
 {
   if rvm_pretty_print stderr
-  then rvm_printf_to_stderr "%b" "${rvm_error_clr:-}$*${rvm_reset_clr:-}\n"
-  else rvm_printf_to_stderr "%b" "$*\n"
+  then printf "%b" "$(__rvm_replace_colors "<error>$*</error>\n")" >&6
```
mpapis: double `printf "%b"` it should be enough to:
```bash
then __rvm_replace_colors "<error>$*</error>\n" >&6
```
and this happens in few places bellow too

pkuczynski: Yes, but then would ` >&6` work well?

PR #4051, opened 2017-05-27
Fix match MacPorts in non standard location

graffic:
There were two problems with the match:
- The start in the beginning would match the entire path
- The parenthesis do match a parenthesis and not the path.


scripts/functions/requirements/osx_port
```diff
@@ -144,7 +144,7 @@ requirements_osx_port_after()
   [[ -n "${port_root}" && -x "${port_root}"  ]] && # verify binary and
   __rvm_readlink_deep port_root                    # resolve any links to point to the final target
   [[ -n "${port_root}" && -x "${port_root}"  ]] && # verify binary and
-  port_root="${port_root%%*(/)bin*(/)port}"        # get    root
+  port_root="${port_root%%/bin/port}"              # get    root
```
mpapis: the `*(/)` was supposed to match duplicated `/` in links... but his version should work just fine in most cases ;)

PR #4043, opened 2017-05-19
Keep all version meta and methods in one place

pkuczynski:
Minor code cleanup:

* Keep all version meta and methods in one place
* Move `rvm_installed_at` to a separate file and rename to `__rvm_version_installed_when`
* Move `versions` file under /scripts/functions

PR #4054, opened 2017-05-31
Give clarity to code block installed to dot-files

### LINKED ISSUE #4053: Add clarity to dotfile edits
## Description

The RVM installer edits several dot-files (and creates them, if they do not already exist). It can be confusing for a user to find dot-files in their home directory for shells they have never installed (e.g. .mkshrc, if they are not a mksh user). It can also be hard for a novice user to know which edits to their dot-files were made by what, when, and why.

Possible example of such confusion: https://github.com/rvm/rvm/issues/3130

## Steps to reproduce

 1. Install RVM according to the instructions.
 2. Several dotfiles (`.mkshrc`, `.bashrc`, `.profile`, `.bash_profile`, etc) are created or modified.
 3. The user might not, at some point in the future, know exactly which ones were modified by RVM, when, and how.

## Expected behavior

RVM should clearly delineate the edits it makes to dot-files.

## Actual behavior

RVM is terse. It is hard to be confident whether the file was modified by RVM itself, or did I instead make the edit manually at some point? Why do I even have a `.mkshrc`, given that I don't have `mksh` installed? Example, after installing a slightly earlier version of RVM:

```$ cat ~/.mkshrc 

export PATH="$PATH:$HOME/.rvm/bin" # Add RVM to PATH for scripting
```

## Environment info

N/A

sampablokuper:
Fixes #4053 .

The RVM installer edits several dot-files (and creates them, if they do not already exist). It can be confusing for a user to find dot-files in their home directory for shells they have never installed (e.g. `.mkshrc`, if they are not a `mksh` user). It can also be hard for a novice user to know which edits to their dot-files were made by what, when, and why.
    
This PR fixes that confusion, by noting, for each edit:
- that the RVM installer made the edit
- when the RVM installer made the edit
- the extent of the edit, using octothorpes as delimiters. Layout loosely inspired by examples such as:
    - https://github.com/syl20bnr/spacemacs/blob/master/core/templates/.spacemacs.template
    - https://github.com/nicksp/dotfiles/blob/master/shell/bash_profile

### DIFF
--- a/scripts/functions/installer
+++ b/scripts/functions/installer
@@ -1100,8 +1100,15 @@ setup_user_profile_rc()
     do
       touch "$profile_file"
       printf "%b" "
+###############################################################################
+# At $(date --iso-8601=s), the RVM installer added this block (and also created
+# this file if it did not already exist).
+#########
+#
 # Add RVM to PATH for scripting. Make sure this is the last PATH variable change.
 export PATH=\"\$PATH:$local_rvm_path/bin\"
+#
+###############################################################################
 " >> "$profile_file"
     done
   else
--- a/scripts/functions/installer
+++ b/scripts/functions/installer
@@ -1105,7 +1105,7 @@ setup_user_profile_rc()
 # this file if it did not already exist).
 #########
 #
-# Add RVM to PATH for scripting. Make sure this is the last PATH variable change.
+# Add RVM to PATH for scripting. Ensure this is the last PATH variable change.
 export PATH=\"\$PATH:$local_rvm_path/bin\"
 #
 ###############################################################################
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Simplify and cleanup of `rvm help` output [\#4029](https://github.com/rvm/rvm/pull/4029)
 * Add support for Kali Linux (based on Debian) [\#3958](https://github.com/rvm/rvm/issues/3958) 
 * Railsexpress patches for 2.4.0, 2.4.1 and 2.4-head [\#4050](https://github.com/rvm/rvm/pull/4050)
+* Improve clarity of dotfile edits [\#4053](https://github.com/rvm/rvm/issues/4053)
 
 #### Bug fixes:
 * Use actual executable test instead of mount|grep noexec for robust noexec detection [\#3933](https://github.com/rvm/rvm/pull/3933)


pkuczynski:
This is only for ~/.bash_profile. How about the other files we modify?

scripts/functions/installer
```diff
@@ -1100,8 +1100,15 @@ setup_user_profile_rc()
     do
       touch "$profile_file"
       printf "%b" "
-# Add RVM to PATH for scripting. Make sure this is the last PATH variable change.
+###############################################################################
+# At $(date --iso-8601=s), the RVM installer added this block (and also created
```
pkuczynski: I would probably add RVM version in here too: `RVM 1.29.1 installer`...

scripts/functions/installer
```diff
@@ -1100,8 +1100,15 @@ setup_user_profile_rc()
...
+# this file if it did not already exist).
+#########
+#
+# Add RVM to PATH for scripting. Ensure this is the last PATH variable change.
```
pkuczynski: This line could be written in a more clear way, don't you think? Maybe "Include RVM in the PATH. Ensure it is always at the end of the PATH".

sampablokuper:
@pkuczynski wrote:
> This is only for ~/.bash_profile.
I don't think so. It should apply to all dotfiles in the `search_list`.

pkuczynski:
Indeed, had a look on the too small part of the file, sorry...

pkuczynski:
Waiting for @mpapis review before merge...

scripts/functions/installer
sampablokuper: > I would probably add RVM version in here too: `RVM 1.29.1 installer`

Sounds good. Will push another commit shortly.

scripts/functions/installer
sampablokuper: Agreed. Will push shortly.

scripts/functions/installer
```diff
@@ -1101,11 +1101,12 @@ setup_user_profile_rc()
       touch "$profile_file"
       printf "%b" "
 ###############################################################################
-# At $(date --iso-8601=s), the RVM installer added this block (and also created
-# this file if it did not already exist).
+# At $(date --iso-8601=s), this code block (and also this file, if it did not
+# already exist) was created by:
+# $__rvm_version
```
pkuczynski: I somehow liked more when it said "RVM 1.29 installed" :)

sampablokuper: I agree :) I couldn't immediately see an easy, robust way to achieve that, though. If you know a better mechanism than using `$__rvm_version`, let me know :)

pkuczynski: Did you try ${rvm_version}?

If that does not work we might need to wait until my PR #4043 is reviewed by @mpapis and merged.

sampablokuper: Thanks. Edit pushed.

scripts/functions/installer
```diff
@@ -1100,8 +1100,15 @@ setup_user_profile_rc()
     do
       touch "$profile_file"
       printf "%b" "
-# Add RVM to PATH for scripting. Make sure this is the last PATH variable change.
+###############################################################################
+# At $(date --iso-8601=s), the RVM ${rvm_version} created
```
pkuczynski: Have you tried if that works?

sampablokuper: Not specifically. Any reason why it wouldn't?
```
$ echo $rvm_version

$ rvm_version='1.2.3'
$ echo "At $(date --iso-8601=s), the RVM Installer from RVM ${rvm_version} created"
At 2017-05-31T20:49:11+0100, the RVM Installer from RVM 1.2.3 created
```

sampablokuper: I'm afraid I'm not familiar with these steps that would seem necessary for fully testing:
- Uninstalling RVM (probably easy, but maybe time-consuming)
- Installing RVM from the branch this PR is based on (trickier)

I'm afraid I can't really put any more time into the matter at the moment either, sorry :( I hope that what I've contributed so far is helpful and can make it into RVM once it has been tested/approved/tweaked by you and @mpapis .

Thanks for maintaining RVM and for your great feedback on this PR :)

pkuczynski:
@mpapis any feedback on this?

scripts/functions/installer
```diff
@@ -1100,8 +1100,15 @@ setup_user_profile_rc()
...
+# this code block (and also this file, if it did not already exist).
+#########
```
mpapis: no need for this separator :)

sampablokuper: The idea with the separators is this:

```
########################### <- First long separator indicates start of hunk inserted by rvm
# rvm should put explanation here
### <- Short separator visually separates end of explanation from start of script snippet
rvm should put snippet here
########################### <- Second long separator indicates end of hunk inserted by rvm
```

pkuczynski: I think it's ok to have the short separator.

pkuczynski:
@mpapis in which other places you would like something like this to be added?

mpapis:
`scripts/functions/installer` in the `setup_user_profile_login` just after `rvm_out "just after Adding rvm loading line to ${target_login[*]}."`

there are other implications, when we run with `--auto-dotfiles` it will accumulate the comments, please try: `./install --auto-dotfiles` to check it locally.

pkuczynski:
Indeed multiple runs duplicate the info entry. I am working on a solution for this issue.

pkuczynski:
@mpapis this is finally done I believe! I reversed some of the changes to ensure backward compatibility with the entries added by the previous versions of RVM. Also refactored and simplified some code.

pkuczynski (dismissed):

scripts/functions/installer
```diff
@@ -1019,150 +1021,175 @@ setup_user_profile_check()
 setup_user_profile_detect()
 {
   etc_profile_file="/etc/profile.d/rvm.sh"
-  search_list_mksh=( "$HOME/.mkshrc" "$HOME/.profile" )
-  search_list_bash=( "$HOME/.bashrc" "$HOME/.bash_profile" "$HOME/.bash_login" )
-  search_list_zsh=( "${ZDOTDIR:-${HOME}}/.zshenv" "${ZDOTDIR:-${HOME}}/.zprofile" "${ZDOTDIR:-${HOME}}/.zshrc" "${ZDOTDIR:-${HOME}}/.zlogin" )
```
mpapis: Any particular reason for limiting the list, my idea with all the files was that user can have sourcing/PATH line in any of them as long as it solves their needs, then the code was detecting which entries were already set up and not installing needlessly in other file which we think is more appropriate for it.

scripts/functions/installer
```diff
@@ -1019,150 +1021,175 @@ setup_user_profile_check()
...
   \typeset etc_profile_file profile_file local_rvm_path local_rvm_path_sed
   local_rvm_path="${rvm_path/#$HOME/\$HOME}"
   local_rvm_path_sed="\\${local_rvm_path//\./\\.}"
   local_rvm_path_sed="${local_rvm_path_sed//\//\/}"
 
-  setup_user_profile_detect
-  setup_user_profile_rc
-  setup_user_profile_login
+  if
+    (( rvm_auto_dotfiles_flag == 1 ))
```
mpapis: Does it mean that we will not add the sourcing/PATH lines automatically on fresh install if user does not pass `--auto-dotfiles` flag? This can increase amount of dissatisfied users with not working setup because of missing this lines.

scripts/functions/installer
```diff
@@ -1019,150 +1021,175 @@ setup_user_profile_check()
...
-        __rvm_grep \..*scripts/rvm "$profile_file" >/dev/null
-      then
-        found_login+=( "$profile_file" )
-      elif
-        __rvm_grep source.*scripts/rvm "$profile_file" >/dev/null
+        __rvm_grep "RVM BEGIN" "$file" >/dev/null
       then
-        found_login+=( "$profile_file" )
+        rvm_debug "Removing RVM info line from ${file}"
+        __rvm_sed_i "$file" -e '/RVM BEGIN$/,/RVM END$/d'
```
mpapis: :+1:

scripts/functions/installer
pkuczynski: The different list was used for searching and different for adding. But then they were merged anyway. So I simplified this and use only one list now.

scripts/functions/installer
pkuczynski: Previously condition for adding was:

```if (( rvm_auto_dotfiles_flag == 1 || ${#found_rc[@]} == 0 ))```

so I felt that we are not making a change in functionality, but you are right. It's mistake. I will remove the `rvm_auto_dotfiles_flag` as in the new code we re-add the lines on every install, as I think it simplifies the code a lot.

scripts/functions/installer
mpapis: but I can not find list of the previous files :(

scripts/functions/installer
mpapis: if we keep updating the line at every install it might mess up things? what if user adds something that depends on RVM and we move our code after his addition?

saizai:
One note: I've occasionally found duplicate rvm lines in dotfiles. Would be nice if it cleaned up old ones, or if not doing differently just tweaked them with a "rechecked at date X, nothing to change since added at date Y".

scripts/functions/installer
martinpwagner: In newer UNIXes aren't used anymore

itimkub:
> One note: I've occasionally found duplicate rvm lines in dotfiles. Would be nice if it cleaned up old ones, or if not doing differently just tweaked them with a "rechecked at date X, nothing to change since added at date Y".

PR #4062, opened 2017-06-12
Fixes ruby version detection for openssl compat detection

### LINKED ISSUE #4057: Ruby < 2.4 fail to build on Fedora 26
## Description

Fedora 26 introduced openssl 1.1 whose API is not compatible with 1.0

## Steps to reproduce

1. Default Fedora 26 installation.
2. rvm install 2.2.6

## Expected behavior

Build fails because of a lot of openssl related build fails.

## How to Fix

For backward compat there is compat-openssl10-devel.

So we should:
if Fedora 26 && ruby < 2.4
   compat-openssl10-devel
else
  openssl-devel

After doing the change (unconditionally) the build works fine for me and I can use 2.2.6

SebastianPoehn:
Fixes #4057

Signed-off-by: Sebastian Poehn <sebastian.poehn@sophos.com>

Fixes # .

*
*
*

### DIFF
--- a/scripts/functions/requirements/fedora
+++ b/scripts/functions/requirements/fedora
@@ -29,8 +29,8 @@ requirements_fedora_define_libyaml()
 
 requirements_fedora_define_openssl()
 {
-  case "$1" in
-    (ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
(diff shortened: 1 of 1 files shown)


pkuczynski:
This PR would not work because `${rvm_ruby_version}` shows only pure version without specifying ruby interpreter. It means it might conflict in the future with other Ruby interpreters like `ree`, `rbx` and others.

PR #4065, opened 2017-06-15
Simplify installer output and some cleanup

pkuczynski:
* removed `thank_you` message as it only pollutes important output
* removed bounty source call out
* removed note about searching for help on rvm.io or twitter. github is our main problem reporting channel


scripts/notes
```diff
@@ -8,18 +8,6 @@ source "$rvm_scripts_path/functions/logging"
 source "$rvm_scripts_path/functions/support"
 source "$rvm_scripts_path/functions/utility"
 
-rvm_warn "\
-In case of problems: ${rvm_notify_clr:-}https://rvm.io/help ${rvm_warn_clr:-}and ${rvm_notify_clr:-}https://twitter.com/rvm_io"
```
mpapis: maybe something like:

    In case of problems open a ticket: https://github.com/rvm/rvm/issues

it does not have to be obvious to everybody

pkuczynski: I am not sure if we want even more issues from inexperienced users than we have now and barely handle. Experienced users would make reports much more clear and might provide helpful input - they would know where to report.

Such message can be also added to the rvm.io, instead of being part of the installer. I have not seen any software which would display such stuff during installation.

PR #4096, opened 2017-07-08
Update gem-wrappers to 1.3.0

mpapis:
Update gem-wrappers to 1.3.0: Show warnings when the target binary is missing or is not executable [gem-wrappers \#9](https://github.com/rvm/gem-wrappers/issues/9)


mpapis:
might need another update for #4104

PR #4120, opened 2017-08-12
Fix parse error in cd script

emersonmx:
When I open a new terminal I get the following message:
/home/emersonmx/.rvm/scripts/cd:14: defining function based on alias `cd'
/home/emersonmx/.rvm/scripts/cd:14: parse error near `()'

I don't know why this happen now, because last week everything was working.
Anyway, everything is fine now with this :)

* Remove parse error message when open a new terminal

### DIFF
--- a/scripts/cd
+++ b/scripts/cd
@@ -11,9 +11,9 @@ case "${rvm_project_rvmrc:-1}" in
   [[ -n "${ZSH_VERSION:-}" ]] &&
   __rvm_version_compare "$ZSH_VERSION" -gt  4.3.4 ||
   {
-    cd()    { __zsh_like_cd cd    "$@" ; }
-    popd()  { __zsh_like_cd popd  "$@" ; }
-    pushd() { __zsh_like_cd pushd "$@" ; }
+    function cd()    { __zsh_like_cd cd    "$@" ; }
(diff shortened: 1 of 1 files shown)


m0n9oose:
Here's the answer

> Seems like a new safe-guard
avoiding redefining an alias as function without unaliasing it first.

https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1539845.html

So you just replace alias definition with function. It's not a solution, and may have unexpected side effects.

emersonmx:
@m0n9oose So it was just a missing setopt...
Thanks I will close this pull request now.

lexicalunit:
@emersonmx what setopt did you change to work around this issue? I'm seeing this same thing:

```
$ source .rvm/scripts/rvm
/Users/me/.rvm/scripts/cd:14: defining function based on alias `cd'
/Users/me.rvm/scripts/cd:14: parse error near `()'
```

Note that changing the line to:

```shell
    function cd()    { __zsh_like_cd cd    "$@" ; }
```

Seems to resolve the issue. The reason `cd` is an alias in the first place is due to https://github.com/sorin-ionescu/prezto/blob/master/modules/utility/init.zsh

emersonmx:
@lexicalunit I just put ALIAS_FUNC_DEF in .zshrc

PR #4146, opened 2017-09-03
Add patches for ruby 2.3/4 compilation on GCC 7

mpapis:
fix #4069, fix #4080, fix #4115

PR #4165, opened 2017-09-15
Add missing random.c patch for Ruby 2.3.2

joesiewert:
Getting the following message when trying to build Ruby 2.3.2:

```
ruby-2.3.2 - #extracting ruby-2.3.2 to /home/rof/.rvm/src/ruby-2.3.2....
ruby-2.3.2 - #applying patch /home/rof/.rvm/patches/ruby/ruby_2_3_gcc7.patch.
Patch 'random_c_using_NR_prefix' not found.
There has been an error applying the specified patches. Halting the installation.
```

It looks like `random_c_using_NR_prefix.patch` was added in #4048, but did not get included for 2.3.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 #### Bug fixes:
 * ZSH Bad pattern for Gemfile ruby declaration [\#4154](https://github.com/rvm/rvm/issues/4154) [\#4156](https://github.com/rvm/rvm/issues/4156)
+* Add missing random.c patch for Ruby 2.3.2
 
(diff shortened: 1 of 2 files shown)


pkuczynski:
@mpapis doesn't it have to be also added to https://github.com/rvm/rvm/blob/master/patchsets/ruby/2.3.2/default ?

mpapis:
it is in [default](https://github.com/rvm/rvm/blob/master/patchsets/ruby/2.3.2/default) already, it was added in #4146

pkuczynski:
Ah, my bad, sorry!

PR #4177, opened 2017-09-19
Require libdb-4.8 for OpenSuSE Tumbleweed

alexandru-calinoiu:
Fixes https://github.com/rvm/rvm/issues/4178

* Use the proper version of libdb in OpenSuSE tumbleweed


pkuczynski:
@mpapis I am wondering if we even need the `_system_name` check here?

mpapis:
yes, `opensuse` handles both SUSE and OpenSUSE... but then we could drop the check with `*` in front ;)

pkuczynski:
I know it does handle both, but after this change, this part of condition will be always true :)

noraj1337:
The last RVM stable is still requiring libdb-4_5 instead of  libdb-4_8.

pkuczynski:
Have you tried `rvm get master` and then install?

PR #4183, opened 2017-09-21
Update the JRuby 1.7.19 checksums

iconara:
As you probably know, all JRuby releases were recently lost (see https://github.com/jruby/jruby/issues/4789) and the JRuby team has been scrambling to restore binaries over the last week. Most releases could be recovered from Maven Central, but unfortunately 1.7.19 seem to have gotten different checksums when released to Maven compared to what was previously in JRuby's own repository.

@enebo asserts in https://github.com/jruby/jruby/issues/4789#issuecomment-330999450 that the file currently at https://s3.amazonaws.com/jruby.org/downloads/1.7.19/jruby-bin-1.7.19.tar.gz is an actual release of JRuby 1.7.19.

The checksums in this commit have been obtained by downloading that file and running `openssl dgst -sha512 FILE` and `openssl dgst -md5 FILE`.


pkuczynski:
I like your comment `I would not approve a PR like that` in jruby/jruby#4789  :)

We are here to help! But this time I am not sure. @mpapis what you think? Maybe we should wait for @enebo?

iconara:
@pkuczynski I was wondering how you would react 😄 I completely understand if you don't feel comfortable merging this.

mpapis:
It might be easier to drop the the checksums and update it once we are sure it's the one to stay.

For JRuby there are `.md5` files online but adding check of them would be an additional effort.

enebo:
@mpapis I think if for some reason I don't have proper backups then perhaps this change might make sense to not impact users as much (although they would need to update rvm to see these new signatures).  The fall back plan I think would be to make a point release 1.7.19.1 which is literally rebuilding 1.7.19 with only the version changed.   @iconara if you want me to do this we can at least get a workaround in place for you.

It is really unfortunate that S3 crapped out.  Our biggest mistake was thinking buckets were set up with restore (we did not set this s3 env up).  I still hold out that next weds when I get home these will all be on my NAS, but this makes me super nervous :|

iconara:
@enebo a 1.7.19.1 release would be great. I will still keep my fingers crossed that you find the old 1.7.19 release too, but a new release would work for the things we're actively working now. We have around 20 repos that use 1.7.19, but most of them aren't in active development. I'm mostly worried about the long term, we're not going to remember this in six months and will probably spend hours if not days debugging why our builds don't work on Travis when we need to make changes to one of the less actively developed code bases.

mpapis:
Let's wait with any changes till @enebo can confirm the file is recoverable or not, we can then push updated md5s or just leave it as it is.

enebo:
@iconara @mpapis well looking back on my old machine (not even my backups and I have a 1.7.19 there) so original bits are stored.  Closing this.

enebo:
Err...this is not my project :P  Someone can close it though

pkuczynski:
Thanks @enebo!

PR #4213, opened 2017-10-27
Change clang version from 3.5 to last available for rbx in Debian

### LINKED ISSUE #4212: Rubinius Ubuntu dependency clang
On Ubuntu 17.04 there is no clang-3.5, available is 3.7+

Both `rvm install rbx` and `rvm install rbx-2.1.1` fail. RVM version is 1.29.2

Thanks.

rtrv:
Fixes #4212


pkuczynski:
How about older version of Debian which would not have clang 3.7 available? You would need to figure out when 3.5 has been stoped and use it conditionally

rtrv:
@pkuczynski thanks, didn't think about it. How about just not fixing the version? I've updated PR

pkuczynski:
Would that work for any version of Debian and rbx? Maybe old rbx would require old version of clang?

mpapis:
rubinius has requirements in regards to the supported versions of clang, it is required to check each version in combination with each system to see what needs to change when - I know, this will be time consuming

pkuczynski:
Close in favour of #4325

PR #4264, opened 2017-12-18
railsexpress patches for ruby trunk, 2.2.9, 2.3.6, and 2.4.3

skaes:
I'd like to support railsexpress patches for ruby head.


skaes:
Added patches for recently released ruby versions.

pkuczynski:
Thanks!

PR #4267, opened 2017-12-26
Set recommended MRI version to 2.5.0

sikachu:
Follow-up from #4265.

This will make RVM install Ruby 2.5.0 when user do `rvm install ruby`.


koic:
@sikachu Thanks for your follow-up 🙏

PR #4270, opened 2017-12-27
Fix fd_mask detection on OS X for Ruby 2.5.0

skaes:
See https://bugs.ruby-lang.org/issues/14248


skaes:
This is just for 2.5.0. I hope they'll fix trunk soon.

PR #4284, opened 2018-01-16
Add prelude_gcc_diagnostic patch for Ruby 2.5.0 to correct compilation failures with GCC < 5

nbeyer:
This is a patch that's already been accepted into ruby-trunk. The issue is here,
https://bugs.ruby-lang.org/issues/14234. The change to ruby-trunk is here,
https://github.com/ruby/ruby/commit/d0afbff3b047755f6a74446185cf08b5618387a0.


pkuczynski:
Thanks!

PR #4300, opened 2018-02-08
Correctly look for rvm group in /etc/group.

### LINKED ISSUE #4296: Multi-user install problem when user or group name contains rvm
When the system already has a user or an group with `rvm` in its name, the following problem occurs:

```
rvmadmin@locahost:~$ \curl -sSL https://get.rvm.io | sudo bash
Downloading https://github.com/rvm/rvm/archive/master.tar.gz
Installing RVM to /usr/local/rvm/
chown: invalid group: ‘:rvm’
    could not set group of '/usr/local/rvm/man/man1', prefix the command with 'rvmsudo' to fix it, if the situation persist report a bug.
chown: invalid group: ‘:rvm’
    could not set group of '/usr/local/rvm/man/man1/rvm.1', prefix the command with 'rvmsudo' to fix it, if the situation persist report a bug.
chown: invalid group: ‘:rvm’
    could not set group of '/usr/local/rvm/man/man1/rvm.1.gz', prefix the command with 'rvmsudo' to fix it, if the situation persist report a bug.
test -f /etc/profile.d/rvm.sh && source /etc/profile.d/rvm.sh
chown: invalid group: ‘:rvm’
    could not set group of '/etc/rvmrc', prefix the command with 'rvmsudo' to fix it, if the situation persist report a bug.
cat: /usr/local/rvm/config/alias: No such file or directory
chown: invalid group: ‘root:rvm’
find: `rvm' is not the name of an existing group
Installation of RVM in /usr/local/rvm/ is almost complete:

  * First you need to add all users that will be using rvm to 'rvm' group,
    and logout - login again, anyone using rvm will be operating with `umask u=rwx,g=rwx,o=rx`.

  * To start using RVM you need to run `source /etc/profile.d/rvm.sh`
    in all your open shell windows, in rare cases you need to reopen all shell windows.
```

I think this is because `__rvm_group_exists()` uses `"$1"` (i.e. `"rvm"`) as regular expression when greping `/etc/group`. A corrected one could be `"^$1:"` (i.e. `"^rvm:"`).

I can't provide a patch because I don't understand how to test it locally.

malin-as:
Fixes #4296 .

* Use an improved regular expression to avoid matching group names that contain the string `rvm`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 * Fix installing rubygems 2.7 [\#4287](https://github.com/rvm/rvm/pull/4287)
 * Add prelude_gcc_diagnostic patch for Ruby 2.5.0 to correct compilation failures with GCC < 5
 * Fix installing rubygems master [\#3734](https://github.com/rvm/rvm/issues/3734)
-* Correctly look for rvm group in /etc/group
+* Correctly look for rvm group in /etc/group [\#4300](https://github.com/rvm/rvm/issues/4300)
 
 #### Upgraded Ruby interpreters:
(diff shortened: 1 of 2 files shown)


pkuczynski:
Thanks!

PR #4314, opened 2018-02-22
Capturing grep for #ruby= in variable to handle error

davisben85:
Capturing the return from __rmv_grep in a variable, and checking that, rather than letting it return a non-zero exit code.  The previous way would fail a bash script if set -e was enabled.   This way, we continue on failure with a non-zero exit code, but still check if we find a file with the #ruby in it.


davisben85:
The issue that this reaolves is when you set up source in your .bash_profile, then try to change directories while using set -e to exit a script as soon as a non-zero exit code is raised.  Changing directories and then running echo $? returns a non-zero exit code IF no gemfile is found that matches the regex in the code that I wrapped. Grep silently throws an error and because of that, it causes the most basic scripts to fail if you are trying to use set -e. I literally cannot change directories without it erroring and exiting.  I fixed it this way, locally, but would appreciate it if this can be reviewed and included with rvm so I don’t have to hack up my local installs.

davisben85:
A workaround, too, for anyone that might be interested, is to use ```builtin cd``` vs just ```cd``` to not use the hook.

mpapis:
pushed cleaner version, let's wait for tests to see if it works ;)

PR #4329, opened 2018-03-14
Add support for installing Rubinius on Redhat/Fedora

pkuczynski:
Based on: https://book.rubinius.com/manuscript/getting_rubinius.html


scripts/functions/requirements/fedora_yum
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+\#!/usr/bin/env bash
```
mpapis: but why?

pkuczynski: typo :) good catch!

PR #4360, opened 2018-04-12
Add install from tag

### LINKED ISSUE #4312: rvm install --url, --branch undocumented; lacks --tag
<s>1. The ability to install a ruby from a git url + branch w/ --url & --branch isn't documented in e.g. `rvm help install`. It should be.</s> ETA: Fixed in #4350.

2. It should also allow git tags, not just branches, which is how e.g. rubinius marks its releases.

saizai:
Fixes #4312.

* add `--tag` to `rvm install`
* add tag to ruby string


mpapis:
all is fine, but wouldn't it be better to fix `rvm_ruby_tag`?

saizai:
Seems to be two totally different things - `rvm_ruby_tag` is basically
the "version" of a ruby; `rvm_ruby_repo_tag` is tag in the `git` sense
(i.e. a nickname for a particular commit).

There's no guarantee that rubies will be stored in git repos or have
git tags to mark their releases. Nor will all git tags be releases.

I tried to make a minimal change so as to *not* mess with the
`rvm_ruby_tag` code. :-P

scripts/functions/selector_parse
```diff
@@ -46,20 +46,23 @@ __rvm_ruby_string_fuzzy()
...
   branch_name="${rvm_ruby_repo_branch:-}"
   ruby_name="${rvm_ruby_name:-}"
+  tag_name="${rvm_ruby_repo_tag:-}"
 
   __rvm_unset_ruby_variables
 
   rvm_ruby_repo_url="${repo_url:-}"
   rvm_ruby_repo_branch="${branch_name:-}"
   rvm_ruby_name="$ruby_name"
+  rvm_ruby_repo_tag="${rvm_ruby_repo_tag:-}"
```
mpapis: rvm_ruby_repo_tag="${tag_name}"

scripts/functions/manage/base_fetch
```diff
@@ -248,6 +263,8 @@ __rvm_fetch_from_git_revision_or_sha()
   then __from="${rvm_ruby_sha}"
   elif [[ -n "${rvm_ruby_repo_branch:-}" ]]
   then __from="${rvm_ruby_repo_branch}"
+  elif [[ -n "${rvm_ruby_repo_tag:-}" ]]
+  then __from="${rvm_ruby_repo_tag}"
```
mpapis: how this is different from the earlier line `__from="${rvm_ruby_tag#t}"`? why can't we use that?

saizai:
rvm_ruby_tag is the "name" of a release for display purposes. It may or
may not be stored as a git tag in the repo.

rvm_ruby_repo_tag (what I added) is specifically a git tag. It may or may
not be a release.

saizai:
`__from` gets used, here, as `git checkout $__from`. `git checkout` works equally well for tags, branches & commit IDs, so I just reused it.

In other parts, it makes a difference — e.g. the gem name & pulling the tag description from git. Or going around the tag-as-in-name lookup and just doing tag-as-in-git.

PR #4381, opened 2018-05-12
Ruby Code and documentation on RVM and Ruby

vishaal-sridhar:
Added Ruby and RVM documentation.

Added Ruby code to show programs in Ruby.

### DIFF
--- a/example-code.txt
+++ b/example-code.txt
@@ -1,0 +1,66 @@
+Here's how to write a hello world program in ruby:
+
+print "Hello, World!\n"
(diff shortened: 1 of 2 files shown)


mpapis:
I'm sorry but I'm failing to see the need fort those files, if at all I think it should be responsibility of README.md to explain what RVM stands for, and it should not be RVM responsibility to show how to use Ruby.

Please let me know if you can provide more details on what was you intent and how it can help other people.

PR #4406, opened 2018-06-17
Add TruffleRuby to rvm

### LINKED ISSUE #4297: Add support for TruffleRuby
There's some need to integrate interactivity to allow users to accept or decline a license as per [this](https://github.com/oracle/truffleruby/issues/1062).

What's the preferred method for integrating such a license acceptance gate into an RVM Ruby formula?

eregon:
Hello,

Here is a PR to add [TruffleRuby](https://github.com/oracle/truffleruby) to rvm. It fixes issue #4297.
I tried to keep the logic simple, and to update all relevant places mentioning other implementations.

I tested `rvm install truffleruby` locally and it seems to work well.

This is a fully open-source version of TruffleRuby, corresponding to a subset of the Community Edition of GraalVM (see http://www.graalvm.org/downloads/)
It includes the [`native` configuration](https://github.com/oracle/truffleruby#truffleruby-configurations) only, not the `JVM` configuration, so there is no JVM involved.

I have a few questions, which I'll point in the diff, as well as the more general question of should it be considered a "binary ruby", mentioned in https://github.com/rvm/rvm/issues/4297#issuecomment-397843910.

Please review carefully, I'm not familiar with the RVM code base and I can't say I have a very good understanding of it (it's quite large and complex).

cc @mpapis @havenwood


man/man1/rvm.1
```diff
@@ -343,6 +343,11 @@ JRuby, Ruby interpreter on the Java Virtual Machine\&.
 Rubinius
 .RE
 .PP
+\fBtruffleruby\fR
+.RS 4
+TruffleRuby, a high performance implementation of Ruby built on GraalVM\&.
+.RE
+.PP
```
eregon: Note: I updated `rvm.1.gz` with `gzip -k rvm.1`.

scripts/functions/selector
```diff
@@ -383,6 +383,7 @@ __rvm_ruby_string_find()
   else
     __rvm_ruby_string_autodetect
     case "${rvm_ruby_string}" in
+      # Should truffleruby-* be added here?
```
eregon: Should truffleruby-* be added here? It's not clear to me.

scripts/functions/utility_system
```diff
@@ -30,6 +30,8 @@ __rvm_remote_extension()
 # params: ruby-string-to-transform
 __rvm_ruby_package_file()
 {
+  # What's this? Is it just for binary rubies? Do we need to change something for TruffleRuby?
+  # We already set rvm_ruby_package_file in __rvm_select_interpreter_truffleruby().
```
eregon: What's this function?
Is it just for binary rubies?
Do we need to change something for TruffleRuby?
Note that we already set rvm_ruby_package_file in `__rvm_select_interpreter_truffleruby()`.

scripts/mount
```diff
@@ -259,6 +259,7 @@ external_select_name()
   fi
 }
 
+# Should we add an entry for truffleruby here? What's the purpose of this function?
```
eregon: Should I add an entry for truffleruby here?
What's the purpose of this function?
Is it for guessing a good name based on the URL/folder given to `mount`?

`rvm mount -r 'https://github.com/oracle/truffleruby/releases/download/vm-1.0.0-rc2/truffleruby-1.0.0-rc2-linux-amd64.tar.gz'` seems to choose `truffleruby-1.0.0-rc2-linux-amd64` as the name currently.
It's mostly fine, except we might want to drop the `-linux-amd64`.
OTOH, it would then conflict with the normal name from `rvm install truffleruby`, which I'm sure is a good idea.
Maybe it should have some prefix or suffix when `rvm mount`-ed?

eregon:
I think it would be nice to test installing TruffleRuby with rvm in Travis CI.
Should I just add a variant of `rvm-test/long/jruby_comment_test.sh` then?

scripts/functions/selector
mpapis: no need

scripts/functions/utility_system
mpapis: This would make sense if we install TruffleRuby, prepare binary for it, upload it to RVM binaries and then people download them, I'm not sure this would work with TruffleRuby license, but still could be used for multiserver deploys in private environment.

I guess we can add this in separate PR.

scripts/mount
mpapis: see the comment in `scripts/functions/utility_system`

mpapis:
merged in https://github.com/rvm/rvm/commit/70ff3d0af27e55a2b70f66216e5d1919c5bcd840

I have left comments about mounting binary rubies, we can address this in separate PR.

mpapis:
btw. great job!

eregon:
@mpapis Thank you for the review and merging!

chrisseaton:
There should be no licence issue with modifying the TruffleRuby distribution by doing things such as precompiling C extensions. The standalone Ruby distribution contains only open-source code under simple licenses (BSD, GPL, etc).

PR #4410, opened 2018-06-19
Call __rvm_setup_compile_environment() for TruffleRuby

eregon:
* So necessary packages get installed and the per-distribution logic gets triggered.

I missed the fact `__rvm_setup_compile_environment()` is the function triggering the per-distribution logic to install packages in https://github.com/rvm/rvm/pull/4406.

cc @mpapis @havenwood 

Relates to #4297.

### DIFF
--- a/scripts/functions/manage/truffleruby
+++ b/scripts/functions/manage/truffleruby
@@ -13,6 +13,7 @@ https://github.com/oracle/truffleruby/blob/master/doc/user/installing-llvm.md"
 
 truffleruby_install()
 {
+  __rvm_setup_compile_environment "${rvm_ruby_string}" || return $?
   truffleruby_install_check_llvm || return $?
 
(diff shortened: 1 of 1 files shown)


eregon:
Note I reused the existing logic for OpenSSL in #4406, which means for instance on Fedora it forces uninstalling `compat-openssl10-devel` and installing `openssl-devel`.
This is actually stricter than what is needed, we support libssl 1.0.2 and 1.1.0 in TruffleRuby.
I think the same applies for recent MRI.

I guess the logic behind this is more deterministic builds and use the latest available libssl to get the latest security patches. Did I guess right?

eregon:
@mpapis @pkuczynski Could you review this PR?

havenwood:
Looks good. Thank you!

eregon:
@havenwood Thanks for merging!

PR #4411, opened 2018-06-22
Reuse the logic for MRI/Rubinius for TruffleRuby for requirements on macOS

eregon:
* requirements_osx_brew_after() expects some global variables to be set,
  notably $brew_libs, $brew_libs_conf and $brew_openssl_package.

@havenwood told me the logic on macOS for TruffleRuby wasn't quite working as expected.
Now TruffleRuby uses the same logic as MRI/Rubinius, as there are complex dependencies between functions due to global variables used within `requirements_osx_brew_after()`.
This installs a few more packages than needed (`autoconf`, `automake`), but for now this seems the safer and simpler fix.

Relates to #4297.

### DIFF
--- a/scripts/functions/requirements/osx_brew
+++ b/scripts/functions/requirements/osx_brew
@@ -301,6 +301,9 @@ requirements_osx_brew_libs_default()
         (rbx*head|rubinius*head)
             brew_libs_conf+=( llvm )
             ;;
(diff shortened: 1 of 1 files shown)


havenwood:
This fixes the macOS deps. Thank you!

PR #4413, opened 2018-06-25
Add bundler to list of default gems

mpapis:
Bundler is now part of rubygems, it appears that rubygems Gemfile
support will get abandoned in favor of bundler and someday
rubygems and bundler will get merged fully.

Till that time we will get back bundler installing by default.


havenwood:
The future for Bundler seems unclear. Currently, if you install Ruby 2.5 you get RubyGems 2.7.6, without Bundler. Ruby core removed Bundler before the Ruby 2.5 release. If you `gem update --system` to RubyGems 2.7.7, voila, you have Bundler. It's unclear if Bundler will once again be removed from RubyGems 3.0 by Ruby core prior to the Ruby 2.6 release.

> My new favorite way to recommend installing Bundler: `gem update --system`.

I think that, upstream, Bundler should either be removed from RubyGems or accepted into Ruby. It seems we need to open avenues of communication between the relevant stakeholders.

PR #4414, opened 2018-06-27
Add jruby-9.2 alias

perlun:
Can we have both this and `jruby[-9.2.0.0]` or will they cause conflicts?

(I can add the CHANGELOG entry if you are fine with the change, change was made right in the web UI so sorry for it not being present already.)

### DIFF
--- a/config/known
+++ b/config/known
@@ -19,6 +19,7 @@ ruby-head
 jruby-1.6[.8]
 jruby-1.7[.27]
 jruby-9.1[.17.0]
+jruby-9.2[.0.0]
 jruby[-9.2.0.0]
 jruby-head
 


mpapis:
This does not make much sense, this strings are used only in displaying `rvm list known` - the square brackets are denoting possible shortening, but RVM is for a long time able to use shorter names and detect things like `rvm install jruby-9.2` or even `rvm install 9.2`.

Closing this ticket, please reopen and let me know if there is some bigger reasoning behind it.

perlun:
@mpapis Sorry for the confusion. I noticed in a PR/Travis build (https://travis-ci.org/ruby-git/ruby-git/jobs/396541130, https://github.com/ruby-git/ruby-git/pull/371) that `rvm install jruby-9.2` didn't work - hence, this was my attempt to fix it.

Now, the root cause of that can have been the JRuby S3 image or whatever. If you say `rvm install jruby-9.2` works now, without this change, I'm perfectly fine with killing it.

PR #4419, opened 2018-07-02
Add TruffleRuby 1.0.0-rc3

eregon:
TruffleRuby 1.0.0-rc3 [has been released](https://github.com/oracle/truffleruby/releases).
In general, we have a release about once a month.

I also updated CONTRIBUTING.md, since the documentation for `config/known_strings` was wrong.
Ideally, we would also link to a different commit than https://github.com/rvm/rvm/commit/c5845cf75f030f8e881e6ab3554dee4f9cc72b46, as that removes a line from `known_string` and I think that's not correct anymore (in master, all versions are in `known_strings`).

cc @havenwood


eregon:
https://github.com/rvm/rvm/commit/f324ed3946e5d74a80c59d873ac346e70dd50612 seems a good example commit for adding support for new version of Ruby, I'll link to that.

PR #4420, opened 2018-07-02
Updates and fixes to switch to new JRuby binary URL.

### LINKED ISSUE #4223: New JRuby download URLs 9.1.14.0+
JRuby's download URLs are changing! We will soon stop hosting our release artifacts on S3 due, and instead would prefer that all downloads come from Maven servers.

To that end, we have started pushing additional checksums with our "jruby-dist" artifact so these download URLs can be considered canonical.

This change affects JRuby 9.1.14.0 and higher. Previous releases will remain on S3 until it goes away.

There are three changes needed in RVM:

The base URL for the binary release tarball changes to `https://repo1.maven.org/maven2/org/jruby/jruby-dist/<version>/jruby-dist-<version>-bin.tar.gz`

The base URL for the source release tarball changes to `https://repo1.maven.org/maven2/org/jruby/jruby-dist/<version>/jruby-dist-<version>-src.tar.gz`

And as you see here, the structure of the tarball file changed from `jruby-bin-<version>.tar.gz` to `jruby-dist-<version>-bin.tar.gz` due to Maven requirements.

We do not know when our S3 will go away, so the sooner you can update how you fetch these files, the better!

headius:
* Switch download URL to Maven address.
* Fix structure of tarball filename.
* Allow "HTTP/2 200" format in curl results.

Fixes #4223.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@
 [Full Changelog](https://github.com/rvm/rvm/compare/1.29.4...HEAD)
 
 #### New features:
+* Switch to new maven-based JRuby download URLs.
 
 #### Bug fixes:
+* Allow HTTP 2.0 servers to be used for downloads.
 
 #### Upgraded Ruby interpreters:
 
--- a/config/db
+++ b/config/db
@@ -5,8 +5,8 @@ niceness=0
 #
 # RVM
 #
-rvm_remote_server_path1=downloads
-rvm_remote_server_url1=https://s3.amazonaws.com/jruby.org
+rvm_remote_server_path1=maven2/org/jruby/jruby-dist
+rvm_remote_server_url1=https://repo1.maven.org
 rvm_remote_server_url2=https://rubies.travis-ci.org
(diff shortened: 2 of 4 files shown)


headius:
Unsure if I did changelog right...let me know if changes are needed.

headius:
I was also not sure if these changes are correct. The downloaded JRuby tarball in archives seems to go to an unusual filename like `bin-jruby-9.2.0.0.tar.gz` for some reason.

headius:
cc @pkuczynski @mpapis for review and assistance :-)

mpapis:
Just checked, works as expected :)

The file name locally is to unify, not all distributions have different name for source and binary, we used the `bin-` prefix to distinguish that - I guess each ruby should have it's own file name builder to allow using original name when possible ... but it would be some refactoring that we are trying to prevent if it's working - unless @pkuczynski got some time to play with it.

headius:
@mpapis Thank you!

ioquatix:
Thanks everyone for sorting this out so quickly!

PR #4427, opened 2018-07-10
Remove early check for LLVM with TruffleRuby

eregon:
* TruffleRuby already checks later and looks up more paths to find the LLVM executables clang and opt.
  (TruffleRuby also looks up in Homebrew and MacPorts directories on macOS).
* See https://github.com/oracle/truffleruby/issues/1386 for details.

So this logic is redundant because per-platform checks for installing LLVM are already in RVM and TruffleRuby itself checks if `clang` & `opt` are available when it needs to compile C extensions.
It's also needlessly restrictive for Homebrew and MacPorts (forcing the LLVM executables to be in `PATH`), so it needs to be removed.

cc @havenwood

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 #### Bug fixes:
 * Allow HTTP 2.0 servers to be used for downloads.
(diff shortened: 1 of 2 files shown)


eregon:
@havenwood Thanks for merging so quickly :)

PR #4428, opened 2018-07-13
UPDATE signals handling

phoolish:
Fixes #4422 on Amazon Linux 2018.03

Per @charlietag comment at https://github.com/rvm/rvm/commit/3d9b93b3ce1a0651c3c7cc23193957f619f65fd9#r29693653:

> Please try modify this line as following:
> ````
> trap >/dev/null
> _rvm_old_traps=$( trap | __rvm_grep -E 'EXIT|HUP|INT|QUIT|TERM' )
> trap >/dev/null
> ````
> Otherwise (trap) will always get the following result:
> trap -- '' SIGINT
> trap -- '' SIGQUIT
> This means.... INT QUIT (__rvm_grep) will always be trapped.
> This means... My ctrl+c will always be blocked....

PR #4432, opened 2018-07-19
Fixed CentOS Ctrl - C trap issue

charlietag:
Fixes #4428 #4422 

* In CentOS `cat <(trap)` will always get the following result:
  `trap -- '' SIGINT` `trap -- '' SIGQUIT`
* This modification could solve Ctrl - c problems in CentOS

### DIFF
--- a/scripts/functions/environment
+++ b/scripts/functions/environment
@@ -212,7 +212,7 @@ __rvm_setup()
     fi
     set +o nounset
 
-    _rvm_old_traps=$( __rvm_grep -E 'EXIT|HUP|INT|QUIT|TERM' <(trap) )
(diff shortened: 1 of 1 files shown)


prioux:
Ah, I was looking at that very line of code last week, and it didn't occur to me that `<(trap)` would have a different behavior than `trap |` ..

Good catch.

charlietag:
@prioux Maybe we are really in the different OS/ bash version. 
Mine is CentOS 7.5.    May I know your OS version, thanks ?

havenwood:
Would this cause a regression by reintroducing the Amazon Linux bug from #4428? If so, we should try to conjure a fix that works for both.

charlietag:
@havenwood 
These two (#4428 ) are duplicated pull requests.  I'm closing this one.

PR #4440, opened 2018-08-03
Add TruffleRuby 1.0.0-rc5

eregon:
* No 1.0.0-rc4 as that was a Graal JavaScript-only release.

cc @havenwood @mpapis


eregon:
The CI fails on macOS, but it seems completely unrelated:
```
$ bundle config
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems.rb:241:in `bin_path': can't find gem bundler (>= 0.a) (Gem::GemNotFoundException)
	from /usr/local/bin/bundle:22:in `<main>'
# failed: status = 127 # was 1
```

PR #4456, opened 2018-09-18
Add support for installing TruffleRuby as a RVM binary ruby

eregon:
TravisCI installs Rubies with the `--binary` flag (see https://github.com/travis-ci/travis-ci/issues/9803#issuecomment-422054386).
So let's support installing TruffleRuby as a RVM binary Ruby.
We still need to run the TruffleRuby post-install hook, but there is other stuff happening anyway like installing gems, etc in `rvm install --binary` or `rvm mount`.
As a side-effect, this PR also fixes `rvm mount` for TruffleRuby in the case the OpenSSL version did not match and it is needed to run the TruffleRuby post-install hook.

I originally asked if TruffleRuby should be considered a source or binary Ruby in https://github.com/rvm/rvm/issues/4297#issuecomment-397843910.
I guess it's mostly binary since everything is precompiled, except the OpenSSL C extension might need to be rebuilt if the OpenSSL version doesn't match the precompiled OpenSSL C extension.
Anyway, I want to keep both source and binary as the binary install seems generally more fragile (for instance, it has a timeout, e.g., https://github.com/travis-ci/travis-ci/issues/5294#issuecomment-165600752).

cc @mpapis @havenwood


config/db
```diff
@@ -8,6 +8,8 @@ niceness=0
 rvm_remote_server_path1=maven2/org/jruby/jruby-dist
 rvm_remote_server_url1=https://repo1.maven.org
 rvm_remote_server_url2=https://rubies.travis-ci.org
+rvm_remote_server_path3=releases/download
+rvm_remote_server_url3=https://github.com/oracle/truffleruby
```
eregon: This is unfortunately an issue, and I'm sure what's the best way to address it:
`rvm_remote_server_url3` is documented as a "user-defined" server URL in https://github.com/rvm/rvm/blob/aa3b553261cf91dcb23a04f2db08824b6ef36aa3/help/prepare.md
It's also used by TravisCI for example (see https://travis-ci.org/eregon/mspec/builds/429941797#L477), which probably means that would override the setting here.

What's a good way to fix this?

nirvdrum:
@mpapis @havenwood Is there anything I can do to help get this PR merged? It's unfortunately blocking us from getting TruffleRuby usable in the Ruby matrix available in Travis.

config/db
havenwood: I think ultimately the existing logic that handles binary downloads isn't extensible and needs to be rewritten. Another issue with `rvm_remote_server_url3` is that `rvm_remote_server_url2` appears to be breaking iteration outside the Travis environment, so this doesn't appear to be working as-is.

It would really be more of a punt, but we could use `rvm_remote_server_url0` for TruffleRuby and switch the iterator to start from zero instead of one after blank.

chrisseaton:
@eregon is away at the moment so he may take a few days to get back to this.

pkuczynski:
@mpapis why don't we merge it straight away?

mpapis:
> @mpapis why don't we merge it straight away?

Waiting for tests to pass.

chrisseaton:
Thanks very much!

eregon:
Thanks for merging :)

I'll add a test later to make sure installing TruffleRuby with RVM doesn't accidentally break. I think we could run that test on each commit and not only in the long tests as it should be fairly fast.

nirvdrum:
I'll also extend my thanks. As a group of volunteers I know it can be hard to align schedules. So, thanks a lot for everyone that was able to pitch in and get this merged!

PR #4479, opened 2018-10-24
Remove unused function __rvm_get_user_shell

pkuczynski:
This function has been moved to `warning_zsh` and is used only there, so we should get rid of it I think...


pkuczynski:
Test failure is not related to this change AFAIK

pkuczynski:
Seems that this function is used somehow, but not sure from where...

PR #4507, opened 2018-11-28
Update rvm-test and test installing TruffleRuby

eregon:
Runs the test added in https://github.com/rvm/rvm-test/pull/26 in rvm/rvm's CI.
Relates to #4297 and https://github.com/rvm/rvm/pull/4456#issuecomment-432487811.

cc @mpapis @pkuczynski @havenwood

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ script:
 env:
   - TEST=rvm-test/fast/* rvm-test-rvm1/*
   - TEST=rvm-test/long/named_ruby_and_gemsets_comment_test.sh
+  - TEST=rvm-test/long/truffleruby_comment_test.sh
 matrix:
   include:
     - os: linux


eregon:
@pkuczynski Thanks for merging!
Unfortunately it seems to fail on Linux as the TravisCI configuration specifies `sudo: false`, while `rvm-test`'s uses the default which is now `sudo: true`.

I think we should change to `sudo: true`, because it's relevant to also test installing the necessary packages, and the container infra is getting deprecated: https://docs.travis-ci.com/user/reference/overview/#for-a-particular-travisyml-configuration

pkuczynski:
Sure. Can u fire up a PR?

eregon:
@pkuczynski See https://github.com/rvm/rvm/pull/4508

Issue #4402, opened 2018-06-07
Cannot install ruby on Mac OS Mojave (10.14 beta)


duhanebel:
when trying to install a version of ruby on Mac OS 10.14 beta 1, the pre-requirments check fails with the following:
```
Searching for binary rubies, this might take some time.
Found remote file https://rubies.[secure]-ci.org/osx/10.13/x86_64/ruby-2.5.1.tar.bz2
Checking requirements for osx.
Installing requirements for osx.
Updating system - please wait
Xcode version older than 4.6.2 installed, download and install newer version from:
    http://connect.apple.com
After installation open Xcode, go to Downloads and install Command Line Tools.
Error running 'requirements_osx_brew_update_system ruby-2.5.1',
please read /Users/[secure]/.rvm/log/1528193314_ruby-2.5.1/update_system.log
Requirements installation failed with status: 1.
```

mjm:
I'm seeing a similar issue to this, but instead on High Sierra with the Xcode 10 beta 1 installed. It seems like the version check for Xcode thinks that Xcode 10 is a lower version than 4.6.2?

mjm:
It looks like this issue is resolved in the master branch, which has had a few updates to the version sorting.

hugazo:
I can confirm that the manual download of CLI for Xcode 10 Beta solved this (I've been failing to install all the afternoon, with rvm and rvenb).

ivan-leschinsky:
Yep, in master branch it has been fixed.
```sh
rvm get master
rvm install 2.5.1
```

pkuczynski:
Thanks for reporting. It's always best to try with `master`. However the above changes have been already released, so I can safely close...

pkuczynski:
@sfsgagi whats wrong with the pkh-info check? I can fix it following your suggestion...

pkuczynski:
Ignore this comment. Just noticed #4532 :)

quantummachina:
> Had the same problem for Mojave
> In master branch it has been fixed.
> 
> ```shell
> rvm get master
> rvm install 2.5.1
> ```

te mereces lo mejor n_n <3

TalayRuk:
Had the same problem for Mojave 10.14.5 
Run this & it's fixed 
rvm get master
rvm install 2.6.3

Thanks a ton

PR #4532, opened 2018-12-17
Fix Command Line Tools version detection on MacOS 10.14.1

sfsgagi:
Improves #4402

* Add check for CommandLineTools that uses newer package name:
pkgutil --pkg-info com.apple.pkg.CLTools_Executables

When XCode is not installed, the __rvm_detect_xcode_version_at_least
check from osx_brew falls back to __CLT_version_at_least 4.6.0 that
tries to detect the version of command line tools. The command it uses:
pkgutil --pkg-info com.apple.pkg.DeveloperToolsCLI is outdated.
That package doesn't exist in Mojave and this change adds another
fallback that uses the new package name.

Before the change any commands that required building ruby was failing
with:
rvm install 2.5.3
...
Xcode version older than 4.6.2 installed, download and install newer version from:
...

Verified that this solved the issue by successfully executing:
rvm install 2.5.3.


pkuczynski:
Thanks!

PR #4542, opened 2018-12-25
Add support for Ruby 2.6.0

koic:
Ruby 2.6.0 has been released.
https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/


ashmaroli:
There seems to be an issue with the upstream's tarball..
https://bugs.ruby-lang.org/issues/15461

pkuczynski:
Will be fixed in 2.6.1. Is the current one causing an issue for u?

PR #4544, opened 2018-12-25
Set Ruby 2.6.0 as the default Ruby version

havenwood:


pkuczynski:
Looks good! Can u update change log?

havenwood:
Changelog updated!

PR #4576, opened 2019-01-15
Add script to automatically register a new TruffleRuby release in RVM

eregon:
This automates adding a new TruffleRuby version in RVM, and avoids copy-paste errors.

Where should I place the script?
For now, I added it at the root of the repo like e.g., `update-remote.sh`.

cc @pkuczynski @havenwood @nirvdrum

### DIFF
--- a/update-truffleruby.rb
+++ b/update-truffleruby.rb
@@ -1,8 +1,8 @@
 require "digest"
 
-raise "Usage: #{$0} VERSION" unless ARGV.size == 1
+raise "Usage: #{$0} VERSION RELEASE_DIRECTORY" unless ARGV.size == 2
 
-version = ARGV[0]
+version, release_directory = ARGV
 
 def replace_line(file, pattern, replacement)
   lines = File.readlines(file)
@@ -26,7 +26,7 @@ insert_after "config/known_strings", /^truffleruby/, ["truffleruby-#{version}\n"
 %w[md5 sha512].each { |algorithm|
   digests = %w[linux-amd64 macos-amd64].map { |platform|
     basename = "truffleruby-#{version}-#{platform}.tar.gz"
-    archive = File.expand_path("../ruby-versions/pkg/#{basename}")
+    archive = "#{release_directory}/#{basename}"
     digest = Digest(algorithm.upcase).file(archive).hexdigest
     "#{basename}=#{digest}\n"
   }
--- a/update-truffleruby.rb
+++ b/update-truffleruby.rb
@@ -1,0 +1,34 @@
+require "digest"
+
+raise "Usage: #{$0} VERSION" unless ARGV.size == 1
+
+version = ARGV[0]
+
+def replace_line(file, pattern, replacement)
+  lines = File.readlines(file)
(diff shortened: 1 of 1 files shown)


eregon:
BTW, is it necessary to update the ChangeLog for TruffleRuby releases? This would remain as a manual step as there is no easy way to predict the PR number.

update-truffleruby.rb
```diff
@@ -0,0 +1,34 @@
...
+
+replace_line "config/db", /truffleruby_version=/, "truffleruby_version=#{version}\n"
+replace_line "config/known", /truffleruby\[/, "truffleruby[-#{version}]\n"
+
+insert_after "config/known_strings", /^truffleruby/, ["truffleruby-#{version}\n"]
+
+%w[md5 sha512].each { |algorithm|
+  digests = %w[linux-amd64 macos-amd64].map { |platform|
+    basename = "truffleruby-#{version}-#{platform}.tar.gz"
+    archive = File.expand_path("../ruby-versions/pkg/#{basename}")
```
pkuczynski: This is script is specific to your own working environment, isn't it? I am looking at this line...

eregon: Yes, it is. I wasn't too keen on downloading the archives once more when I have them locally already. But I can download them here if you think it's better, or expect the files to be in the current directory, which is more flexible.

pkuczynski: I think we could download when not in current directory (or directory passed as a param?)...

pkuczynski: I still keep wondering where we should locate this file... But maybe for now it's OK to keep it where it is...

eregon: Good idea, I now added the directory as a command-line argument.

eregon:
@pkuczynski Could you merge this PR?

havenwood:
On the question of where to put it, I wonder if it'd be worth keeping it top level but renaming it `update-versions.sh` with just a `truffleruby` argument for now? (Similar to ruby-versions' `update.sh`.)

eregon:
I'm not looking forward to generalize the script for all Ruby implementations at this time, I'd rather merge what works and is useful now.

I could rename to `update-versions.rb` and only support `truffleruby`, but that seems slightly misleading for me.

eregon:
@havenwood @pkuczynski Please tell me what changes you want before merging, and then let's merge it :smiley:

havenwood:
Since it's an internal, contributor-facing script I think we can merge it as-is, then move it to `update-versions.rb truffleruby` once we have another engine's update script to add.

havenwood:
Thank you for the PR!

eregon:
Thanks for merging!

pkuczynski:
Sorry, was on holiday...

eregon:
@pkuczynski No worry. I wanted to merge this soon as it allows to simplify these instructions quite a bit:
https://github.com/oracle/truffleruby/commit/1217310872c302407b1786e59fd5a78c719bd20b?short_path=7400bc7 (and of course makes it much easier to update TruffleRuby in RVM).

Issue #4561, opened 2018-12-28
rvm get master fails in TravisCI


eregon:
## Description

`rvm get master` fails in TravisCI, as seen in this example job:
https://travis-ci.org/eregon/mspec/builds/473147581

The `.travis.yml` is minimal:
https://github.com/eregon/mspec/blob/663b50d3ba978d80fb61f529cc08464fba77c752/.travis.yml

This is annoying, because it [prevents](https://travis-ci.org/ruby/mspec/jobs/473108215) using TruffleRuby in TravisCI.

I think it would be worth to add a check in CI to ensure `rvm get master` works within TravisCI.

## Expected behavior

That it works.

## Actual behavior

```
$ rvm get master
Downloading https://get.rvm.io
Downloading https://raw.githubusercontent.com/rvm/rvm/master/binscripts/rvm-installer.asc
Verifying /home/travis/.rvm/archives/rvm-installer.asc
gpg: Signature made Thu 27 Dec 2018 06:38:39 PM UTC using RSA key ID 39499BDB
gpg: Can't check signature: No public key
Warning, RVM 1.26.0 introduces signed releases and automated check of signatures when GPG software found. Assuming you trust Michal Papis import the mpapis public key (downloading the signatures).
GPG signature verification failed for '/home/travis/.rvm/archives/rvm-installer' - 'https://raw.githubusercontent.com/rvm/rvm/master/binscripts/rvm-installer.asc'! Try to install GPG v2 and then fetch the public key:
    gpg2 --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
or if it fails:
    command curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
the key can be compared with:
    https://rvm.io/mpapis.asc
    https://keybase.io/mpapis
NOTE: GPG version 2.1.17 have a bug which cause failures during fetching keys from remote server. Please downgrade or upgrade to newer version (if available) or use the second method described above.
/home/travis/.rvm/scripts/functions/cli: line 243: return: _ret: numeric argument required
The command "rvm get master" exited with 2.
```

eregon:
Here is the `--debug` output but it doesn't give much information: https://travis-ci.org/eregon/mspec/builds/473150635

pkuczynski:
Your issue is related to the fact, that you are missing gpg keys for the new signer (me). add to your travis line:

[...]

PR #4577, opened 2019-01-15
Test TruffleRuby in TravisCI in RVM's .travis.yml

eregon:
As seen in https://github.com/rvm/rvm/issues/4561#issuecomment-450678661, it happens that the CI of RVM passes, yet `rvm: truffleruby` doesn't work in TravisCI.

So this PR adds a test using `rvm: truffleruby` in RVM's `.travis.yml`.
This is exactly what people testing their Ruby library on TravisCI will use, and so tests one important use-case where RVM is used.

cc @pkuczynski

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,17 @@ matrix:
       script:
         - gpg2 --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
         - gpg2 --verify binscripts/rvm-installer.asc binscripts/rvm-installer
+    - language: ruby
+      os: linux
+      env: TEST="Test TruffleRuby in TravisCI"
+      rvm: truffleruby
+      before_install: true
+      install: true
+      before_script: true
+      script:
+        - ruby -v
+        - rake --version
+        - ruby -ropen-uri -e 'puts open("https://rubygems.org/") { |f| f.read(1024) }'
 notifications:
   email:
     recipients:


mpapis:
hmm, we have already 
[`TEST=rvm-test/long/truffleruby_comment_test.sh`](https://travis-ci.org/rvm/rvm/jobs/480017712#L1346) which is using the same command [`rvm install truffleruby`](https://travis-ci.org/rvm/rvm/jobs/480017717#L486)

eregon:
@mpapis Yes, but `truffleruby_comment_test.sh` is running differently with e.g., setup for installing the local RVM, in a `language: c` image, etc, while this new test is exactly what TravisCI users using TruffleRuby use. As #4561 shows, there are differences between the two, and ultimately what should work is this new test, so I'd like to test it here.

BTW, the CI on macOS has been broken for a while, and is unrelated to this PR. But it makes it harder to notice if something regressed.

eregon:
> BTW, the CI on macOS has been broken for a while, and is unrelated to this PR. But it makes it harder to notice if something regressed.

I filed https://github.com/rvm/rvm/issues/4579

eregon:
Could you merge this PR?

mojavelinux:
I'm finding that truffleruby is taking an aweful long time to install in Travis CI. In some runs, almost 3 minutes. Is there anything that can be done in RVM to speed this up?

See https://travis-ci.org/asciidoctor/asciidoctor/jobs/488821301

eregon:
@mojavelinux I think this PR is not the best place to discuss this :)
I have some timings by using RVM's `--trace` on my other laptop.
Please open an issue at https://github.com/oracle/truffleruby/issues and I'll post them there.
This PR does show that on that run of TravisCI it took `169.16s` for `rvm install truffleruby`, which I agree is too much.

@pkuczynski Could you merge this PR?

pkuczynski:
Sure. I am still trying to recover after holidays ;)

PR #4584, opened 2019-01-19
Don't spoil environment with '_system_*' variables

VladRassokhin:

* Don't spoil environment with '_system_*' variables
I find it quite annoying to have strange env variables in CI environment where rvm is installed and loaded.
Another question is why they don't have 'rvm' prefix.

### DIFF
--- a/scripts/functions/environment
+++ b/scripts/functions/environment
@@ -294,6 +294,8 @@ __rvm_teardown()
   fi
 
   __variables_definition unset
+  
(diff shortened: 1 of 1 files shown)


mpapis:
those variables were supposed to not be leaked, but finding why they were is to much hassle, the fix is enough

pkuczynski:
Thanks!

VladRassokhin:
Thanks for fast-merging :)

PR #4605, opened 2019-02-04
Add TruffleRuby 1.0.0-rc12

eregon:
cc @pkuczynski  @havenwood

BTW, could we merge https://github.com/rvm/rvm/pull/4576 too?


eregon:
Also, is it useful/necessary to update the ChangeLog for new TruffleRuby releases? It's one manual step that cannot be easily automated, causes a second CI run, and doesn't seem so useful.

havenwood:
> Also, is it useful/necessary to update the ChangeLog for new TruffleRuby releases? It's one manual step that cannot be easily automated, causes a second CI run, and doesn't seem so useful.

We used to do the release changelog notes all at once, at the time of release. I think @pkuczynski switched to asking for notes per PR so missing changelog updates wouldn't always be blocking release.

One option might be to use a script to generate the changelog at release time from merged PRs. That might be fairly difficult to do unless we change the changelog format. Or we could stop doing release changelogs.

havenwood:
@pkuczynski How do you feel about the release changelog? Worth attempting to generate it at release time, deprecate it, or keep it as as?

pkuczynski:
Format-wise I don't mind having every version as a separate entry. But it's better to have changelog modified with every PR. Otherwise users for `master` would not know what changed. And creating changelog after weeks of development is painful.

pkuczynski:
I see your point for automation and I think we can live without the PR number :)

eregon:
@pkuczynski  Great, I'll make the script add a line in the ChangeLog then :)

PR #4610, opened 2019-02-07
Automatically add a ChangeLog entry with update-truffleruby.rb

eregon:
cc @pkuczynski


eregon:
Rebased, this should pass the Linux CI now.

Issue #4609, opened 2019-02-07
The CI is failing, the rvm-test submodule was removed


eregon:
## Description

The latest build on master, https://travis-ci.org/rvm/rvm/builds/489053382, shows that all tests using `tf` are broken.

Maybe using an older version of `tf` would fix this?

I propose to make `master` a protected branch and enforce all Linux builds to succeed before merging a PR (until macOS builds also work, #4579).
Otherwise, it's just too easy to merge breaking changes.

cc @pkuczynski

pkuczynski:
Hmm, don't know how and when the module was removed. You are right, we should focus on fixing macOS build... if we would only have more time :/

eregon:
@pkuczynski @havenwood 
Now that the CI is green (with https://github.com/rvm/rvm/pull/4615), could we make `master` a GitHub protected branch, so only PRs passing the CI can be merged, and we can avoid cases like the above report issue?

Specifically, I suggest to make TravisCI a required status check with
https://help.github.com/articles/enabling-required-status-checks/
I wouldn't enable `Require branches to be up to date before merging` as that's rather annoying.

This would also make me (and I think others as well) more confident tests do not stop working without noticing.

pkuczynski:
`master` is already a protected branch, I was just over-ruling it. Sorry about that. Will be more careful in the future!

eregon:
Maybe the ` Include administrators` checkbox is good then :smile:

eregon:
I think this was fixed in #4611, so let's close this issue.
TravisCI is now a required check, but it's not currently enforced for administrators.

Issue #4579, opened 2019-01-16
Fixing the CI on macOS, or remove it


eregon:
The CI on macOS has been failing for a long time.
This makes it hard to notice if a new test fails, i.e., it requires to check the build page every time instead of seeing the whole CI as green.

I think the macOS CI should either be fixed, or removed, as currently it's useless and takes more time.

The problem seems to be that there are too many upgrades with Homebrew which take a very long time.
e.g. https://travis-ci.org/rvm/rvm/jobs/480017710

Could you find out what's triggering these updates, most of which don't seem needed, and disable that?

eregon:
`rvm-test/long/named_ruby_and_gemsets_comment_test.sh` refers to an ancient Ruby version, 1.8.7, maybe it's time to update that one to a supported Ruby version?

It seems installing many Homebrew packages start to get installed from
```
Checking requirements for osx.
Do not know how to check/install gcc 'gcc'.
Installing requirements for osx.
Updating system..........
Installing required packages: zlib....
```
and Homebrew goes in a frenzy updating apparently the whole system. Here is one example:
```
==> Upgrading postgis 
==> Installing dependencies for postgis: cfitsio, popt, epsilon, expat, freexl, xz, python, geos, giflib, gmp, isl, mpfr, libmpc, gcc, szip, hdf5, jpeg, jasper, json-c, libxml2, libdap, libtiff, proj, libgeotiff, libpng, libpq, libspatialite, netcdf, openblas, numpy, pcre, freetype, fontconfig, libidn, podofo, glib, pixman, cairo, little-cms2, nspr, nss, openjpeg, qt, poppler, unixodbc, webp, xerces-c, zstd, gdal, icu4c, postgresql, boost, eigen, cgal and sfcgal
```
https://travis-ci.org/rvm/rvm/jobs/490477283

cc @pkuczynski

eregon:
Quick attempt to stop updating/upgrading Homebrew in https://github.com/eregon/rvm/commit/e9a3e43315b09eda362990963171d7a25afe7e49, https://travis-ci.org/eregon/rvm/builds/490494880

eregon:
Note that many tests currently pass on macOS, although they take a very long time to run:
https://travis-ci.org/rvm/rvm/builds/492123234
```
TEST=rvm-test/fast/* rvm-test-rvm1/*
Requirements installation failed with status: 1.
##### Processed commands 621 of 621, success tests 432 of 434, failure tests 2 of 434.

[...]

PR #4615, opened 2019-02-12
Allow failures on macOS until the CI is fixed on macOS

eregon:
* See https://github.com/rvm/rvm/issues/4579

I'd like the build to be green, to avoid regressions (e.g., #4609).

Another option is completely removing tests on macOS, but many tests pass:
https://github.com/rvm/rvm/issues/4579#issuecomment-462966466

cc @pkuczynski

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,13 @@ matrix:
         - ruby -v
         - rake --version
         - ruby -ropen-uri -e 'puts open("https://rubygems.org/") { |f| f.read(1024) }'
+  allow_failures:
+    - os: osx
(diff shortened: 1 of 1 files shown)


eregon:
Given how long they take to run, I'm starting to think removing them might be better...

eregon:
@pkuczynski @havenwood Could you merge this?

havenwood:
It makes sense to merge this until we resolve the underlying issues. We can revert this commit once macOS tests are green.

eregon:
@havenwood Thanks for merging!

PR #4621, opened 2019-02-19
Fix macOS CI

### LINKED ISSUE #4579: Fixing the CI on macOS, or remove it

eregon:
Fixes #4579.

This PR makes the macOS CI on TravisCI green.

For that, it removes very slow (5min/job vs 20-40min) and unreliable calls to Homebrew by using `rvm autolibs read-only`. Without that, RVM would call to Homebrew and update many packages, which would take a very long time and lead to build failures (#4579).
In practice, it is clear TravisCI macOS machines already have all the necessary dependencies to test RVM and install the tested Ruby versions.

I think we should merge this because it's more valuable to test the rest of the logic (everything except `autolibs`) on macOS, than having a failing CI and issues potentially piling up.

cc @pkuczynski @havenwood

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ before_script:
   - ./install
   - git submodule init
   - git submodule update
+  - rvm reload
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then rvm autolibs read-only; fi
 script:
   - tf --text $TEST
 env:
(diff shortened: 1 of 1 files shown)


pkuczynski:
Yeah, I guess it makes sense. @mpapis ?

PR #4629, opened 2019-03-09
Fix wording (this gems -> these gems)

fwolfst:
Tiny wording change. It is unclear at that moment if its one or multiple gems or anything else that made the command fail. The message now is however better than the mix of singular and plural.


pkuczynski:
Thanks!

PR #4643, opened 2019-03-16
Update rbx dependencies for macOS

pkuczynski:
Base on https://book.rubinius.com/manuscript/getting_rubinius.html


havenwood:
This looks right to me, and mirrors the [deps ruby-install uses](https://github.com/postmodern/ruby-install/blob/master/share/ruby-install/rbx/dependencies.txt). 👍

PR #4649, opened 2019-03-24
Add rvm-3.107

saizai:


pkuczynski:
We dont add md5/sha for rubinius

PR #4650, opened 2019-03-24
Use remote .sha512 and .md5 if available (Rubinius)

saizai:
E.g. https://github.com/rubinius/rubinius/releases/download/v3.107/rubinius-3.107.tar.bz2 -> https://github.com/rubinius/rubinius/releases/download/v3.107/rubinius-3.107.tar.bz2.sha512

(But doesn't have https://github.com/rubinius/rubinius/releases/download/v3.107/rubinius-3.107.tar.bz2.md5)


scripts/functions/checksum
```diff
@@ -198,14 +198,19 @@ __rvm_checksum_read()
...
     [[ -n "${_checksum_md5:-}" ]] ||
       _checksum_md5="$(    __rvm_db_ "$rvm_user_path/md5"      "$_name" | \command \head -n 1 )"
+    [[ -n "${_checksum_md5:-}" ]] ||
+      _checksum_md5="$(    __rvm_curl -s -L $_name.md5  | \command \grep -oe "[0-9a-f]{32}"  )"
 
     _checksum_sha512="$(   __rvm_db_ "$rvm_path/config/sha512" "$_name" | \command \head -n 1 )"
     [[ -n "${_checksum_sha512:-}" ]] ||
       _checksum_sha512="$( __rvm_db_ "$rvm_user_path/sha512"   "$_name" | \command \head -n 1 )"
+    [[ -n "${_checksum_md5:-}" ]] ||
+      _checksum_sha512="$( __rvm_curl -s -L $_name.sha512  | \command \grep -oe "[0-9a-f]{128}"  )"
```
mpapis: I wonder if we need a flag to enable/disable this as this will do extra queries to the web ... maybe we can at least check the `name` starts with `http`

saizai: my bash isn't that good. would suggest you edit it for that

pkuczynski: Done!

scripts/functions/checksum
```diff
@@ -195,14 +195,24 @@ __rvm_checksum_read()
 
   for _name in "${list[@]}"
   do
+    # md5
     _checksum_md5="$(      __rvm_db_ "$rvm_path/config/md5"    "$_name" | \command \head -n 1 )"
+
     [[ -n "${_checksum_md5:-}" ]] ||
       _checksum_md5="$(    __rvm_db_ "$rvm_user_path/md5"      "$_name" | \command \head -n 1 )"
 
+    [[ -n "${_checksum_md5:-}" && $_name == http* ]] ||
```
eregon: This is a bug, specifically it will use `curl` when `-n "${_checksum_md5:-}"`, that is when `_checksum_md5` is not empty, i.e, when it's already set from config files, and when `$_name` does not start with `http`. Fix in https://github.com/rvm/rvm/pull/4707

PR #3856, opened 2016-12-30
Don't hard code number of compile threads

bartoszcisek:
RVM can automatically detect number of available threads.

Changes proposed in this pull request:
* remove hardcoded `-j3`

PR #4674, opened 2019-04-22
rvm-installer: remove hardcoded number of jobs

ftab:
`__rvm_setup_compile_environment_flags_threads()` in scripts/functions/build_config auto-detects the number of threads to use

* remove the hardcoded number of jobs from rvm-installer so that it will auto detect and build with as many threads as it can

Related: #3856


pkuczynski:
Thanks!

PR #4696, opened 2019-05-16
Improve CLI error messages/warnings (#4695)

### LINKED ISSUE #4695: UI: more helpful warnings/error messages?
## Description

    $ rvm --version
    rvm 1.29.8 (latest) by Michal Papis, Piotr Kuczynski, Wayne E. Seguin [https://rvm.io]

I propose two changes to point users to their next action if they ran into "trouble":

### rvm install

    $ rvm install
    Can not use or install 'all' rubies.

The "error" message could be pointing into the direction of finding installation candidates

     Can not use or install 'all' rubies. Get a list of installable rubies by calling `rvm list known`.

### autoupdate warning

I actually came to this from following situation

> $ rvm install
> Warning, new version of rvm available '1.29.8', you are using older version '1.29.
> 7'.                                                                               
> You can disable this warning with:    echo rvm_autoupdate_flag=0 >> ~/.rvmrc      
> You can enable  auto-update  with:    echo rvm_autoupdate_flag=2 >> ~/.rvmrc      
> Can not use or install 'all' rubies. 

I propose to change the text into

> Warning, new version of rvm available '1.29.8', you are using older version '1.29.
> 7'. You can use `rvm get` to update (e.g. `rvm get stable`).
> ...

Would such a PR be merged in?

fwolfst:
Provide the user the commands for the typical next steps.

Fixes #4695.

Added CHANGELOG entry. If that is really wished even for tiny changes like this, give me feedback (I will then also update the CONTRIBUTING (ater opening an issue ;) ), as this policy is not mentioned there).

### DIFF
--- a/scripts/functions/cli
+++ b/scripts/functions/cli
@@ -186,9 +186,10 @@ __rvm_cli_autoupdate_version_old()
 __rvm_cli_autoupdate_warning()
 {
   rvm_warn "Warning, new version of rvm available '${online_version}', you are using older version '${rvm_version%% *}'.
-You can disable this warning with:    echo rvm_autoupdate_flag=0 >> ~/.rvmrc
-You can enable  auto-update  with:    echo rvm_autoupdate_flag=2 >> ~/.rvmrc
-You can update  manually     with:    'rvm get' (e.g. 'rvm get stable')."
+You can disable this warning with:   echo rvm_autoupdate_flag=0 >> ~/.rvmrc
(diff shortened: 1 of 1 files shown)


fwolfst:
Btw, these (and others) could probably be split up into two: `rvm_warn`/`rvm_error` for the warning/error message and `rvm_notify` for the user-directed supportive instructions. Or is `rvm_notify` thought to report progress only? In that case we could introduce a `rvm_info`. Time's short, but I can re-PR with `rvm_notify` if that's preferred.

PR #4703, opened 2019-05-26
Fix endless loop on macOS when listing remotes

felixbuenemann:
If the check for Travis CI binaries returned false – as is always the case outside of Travis CI – the code could go into an endless loop, because is was continued before the iterator was incremented.

### DIFF
--- a/scripts/functions/utility_system
+++ b/scripts/functions/utility_system
@@ -146,12 +146,14 @@ __list_remote_all()
   while
     __rvm_db "rvm_remote_server_url${_iterator:-}" rvm_remote_server_url
   do
-    __rvm_include_travis_binaries || continue
+    if
+      __rvm_include_travis_binaries
+    then
+      __rvm_system_path "" "${_iterator}"
+      rvm_debug "__list_remote_all${_iterator:-} $rvm_remote_server_url $rvm_remote_server_path"
 
-    __rvm_system_path "" "${_iterator}"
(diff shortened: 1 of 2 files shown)


scripts/functions/utility_system
```diff
@@ -146,12 +146,12 @@ __list_remote_all()
   while
     __rvm_db "rvm_remote_server_url${_iterator:-}" rvm_remote_server_url
   do
-    __rvm_include_travis_binaries || continue
+    if __rvm_include_travis_binaries; then
```
mpapis: https://github.com/rvm/rvm/blob/master/FORMATTING.md#sparse-form-for-long-statements--more-commands

felixbuenemann: Sorry, I just used the style from the `__rvm_remote_server_path` function directly above in the same file, which used `if ; then`.

I pushed a new version that follows the style guide and also simplified the commit message.

felixbuenemann:
Travis CI failures seem unrelated, I fixed one of them in #4704.

felixbuenemann:
The other test failure seems to be caused by an incompatible config in `/home/travis/.rvmrc` which has `rvm_gem_options='--no-document'` which gets then merged into `install_params+=( $rvm_gem_options )` so it ends up with `--no-ri --no-rdoc --no-document` which can't work with rubygems 1.8.25.

pkuczynski:
Thanks!

PR #4704, opened 2019-05-27
Install missing dirmngr for GPG signature test

felixbuenemann:
This is needed, because the Travis CI apt addon installs packages with
--no-install-recommends and dirmngr is a recommended package for gnupg2.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845720 for discussion.

No Changelog update since this only concerns CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ matrix:
       addons:
         apt:
           packages:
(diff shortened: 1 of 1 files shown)


pkuczynski:
Thanks!

PR #4705, opened 2019-05-27
Filter redundant/incompatible rvm_gem_options

felixbuenemann:
This removes any occurrence of `--no-document`, `--no-ri` and `--no-rdoc` from `rvm_gem_options` when merging with the auto-detected `install_params` to avoid duplicate, deprecated or incompatible options.

This should fix the current test failures for ruby-1.8.7 under Xenial on Travis CI, where the default Travis `~/.rvmrc` contains `rvm_gem_options='--no-ri --no-rdoc --no-document'` breaking rvm gem installs with rubygems 1.8.25 which does not yet support the `--no-document` option.

### DIFF
--- a/scripts/functions/gemset
+++ b/scripts/functions/gemset
@@ -181,7 +181,17 @@ gem_install_force()
   else install_params+=( --no-ri --no-rdoc )
   fi
 
-  install_params+=( $rvm_gem_options )
+  for __gem_option in ${rvm_gem_options}
+  do
+    case "${__gem_option}" in
+      (--no-ri|--no-rdoc|--no-document)
+        # skip
+        ;;
+      (*)
+        install_params+=( "${__gem_option}" )
+        ;;
+    esac
+  done
 
   __rvm_log_command \
     "gem.install.${gem_name}${gem_version:+-}${gem_version:-}" \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@
 #### Bug fixes
 * Fix rvm version validation per project [\#4692](https://github.com/rvm/rvm/pull/4692)
 * Fix endless loop on macOS when listing remotes [\#4703](https://github.com/rvm/rvm/pull/4703)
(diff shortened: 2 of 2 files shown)


felixbuenemann:
While it would be possible to fix the rvm CI failures simply by patching the existing `~/.rvmrc`, I chose to fix this in rvm instead, since it's likely to help other rvm users.

The regex is fairly defensive, to guard against things like `--no-rdoc-foo`, but has no guards at the beginning of the string due to weird differences in how BSD and GNU sed handle matching. However unless some gem requires an absurd flag like `--not-posix-style--no-document` that shouldn't really be a problem.

A valid case that we wouldn't catch is `rvm_gem_options` with embedded quoting around params:

```bash
rvm_gem_options='"--no-rdoc" "--no-ri"'
```

That's agaib pretty unlikely to be used in rvmrc and IMHO not worth the trouble (especially given the limitations and portability of sed's extended REs).

felixbuenemann:
The remaining test failure is fixed by #4704.

scripts/functions/gemset
```diff
@@ -181,7 +181,7 @@ gem_install_force()
   else install_params+=( --no-ri --no-rdoc )
   fi
 
-  install_params+=( $rvm_gem_options )
+  install_params+=( $(echo "$rvm_gem_options" | __rvm_sed -E 's/--no-(document|ri|rdoc)([[:space:]]+|$)//g') )
```
mpapis: try:
```bash
for __gem_option in $rvm_gem_options
do
  case $__gem_option
    (--no-ri|--no-rdoc|--no-document) true ;;
    (*) install_params+=( $__gem_option )
 esac
done
```

felixbuenemann: Thanks, that's much nicer than my solution!

As long as none of the `rvm_gem_options` contain spaces your solution should work fine.

felixbuenemann: Rebased on master and pushed new version with proposed case statement.

pkuczynski:
Good work, thanks!

PR #4707, opened 2019-05-30
Fix checksum check condition to not try url if already found in config files

eregon:
Fix condition to only try `curl` if no checksum is found
* This would cause checksum checks to fail even if checksums exist in the config files, as long as the given URL does not exist.
* This bug was introduced by #4650

Only try to get the checksum from the network after checking the config files (avoiding extra `curl` queries if not needed).

This notably fixes the checksum check for TruffleRuby.

cc @pkuczynski @mpapis

### DIFF
--- a/scripts/functions/checksum
+++ b/scripts/functions/checksum
@@ -201,7 +201,7 @@ __rvm_checksum_read()
     [[ -n "${_checksum_md5:-}" ]] ||
       _checksum_md5="$(    __rvm_db_ "$rvm_user_path/md5"      "$_name" | \command \head -n 1 )"
 
-    [[ -n "${_checksum_md5:-}" && $_name == http* ]] ||
+    [[ -z "${_checksum_md5:-}" && $_name == http* ]] &&
        _checksum_md5="$(__rvm_curl -s -L $_name.md5)"
 
     # sha512
@@ -210,7 +210,7 @@ __rvm_checksum_read()
     [[ -n "${_checksum_sha512:-}" ]] ||
       _checksum_sha512="$( __rvm_db_ "$rvm_user_path/sha512"   "$_name" | \command \head -n 1 )"
 
-    [[ -n "${_checksum_sha512:-}" && $_name == http* ]] ||
+    [[ -z "${_checksum_sha512:-}" && $_name == http* ]] &&
        _checksum_sha512="$(__rvm_curl -s -L $_name.sha512)"
 
     __rvm_checksum_none || return 0
--- a/scripts/functions/checksum
+++ b/scripts/functions/checksum
@@ -201,17 +201,27 @@ __rvm_checksum_read()
     [[ -n "${_checksum_md5:-}" ]] ||
       _checksum_md5="$(    __rvm_db_ "$rvm_user_path/md5"      "$_name" | \command \head -n 1 )"
 
-    [[ -z "${_checksum_md5:-}" && $_name == http* ]] &&
-       _checksum_md5="$(__rvm_curl -s -L $_name.md5)"
-
     # sha512
     _checksum_sha512="$(   __rvm_db_ "$rvm_path/config/sha512" "$_name" | \command \head -n 1 )"
 
     [[ -n "${_checksum_sha512:-}" ]] ||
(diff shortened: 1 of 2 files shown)


scripts/functions/checksum
```diff
@@ -201,17 +201,27 @@ __rvm_checksum_read()
...
 
     [[ -n "${_checksum_sha512:-}" ]] ||
       _checksum_sha512="$( __rvm_db_ "$rvm_user_path/sha512"   "$_name" | \command \head -n 1 )"
 
-    [[ -n "${_checksum_sha512:-}" && $_name == http* ]] ||
-       _checksum_sha512="$(__rvm_curl -s -L $_name.sha512)"
+    __rvm_checksum_none || return 0
+  done
+
+  # Only try to get the checksum from the network after checking the config files
```
pkuczynski: Why don't we keep the previous version but do:

```
[[ -z "${_checksum_md5:-}" && $_name == http* ]] ||		
        _checksum_md5="$(__rvm_curl -s -L $_name.md5)"
```

eregon: I think that's incorrect too, it would run curl if `$_name` does not start with `http`, even if the checksum is already set. My first commit e370ad601fc045f3c886ad84fe2b04fa75af6eef I think does the minimal fix.

But it seems highly desirable to avoid `curl` calls entirely if there are already entries in the config files. Since the code iterates over multiple `_name`, we need to split the loop, so e.g. config files for the second `_name` (e.g., `truffleruby-19.0.0-linux-amd64.tar.gz`) are checked before a URL for the first `_name` (e.g., a URL). I think this approach somewhat concurs with @mpapis's comment: https://github.com/rvm/rvm/pull/4650#discussion_r269320930, i.e., avoiding needless `curl` calls.

I also find a pair of `if` a lot more readable and easier to reason about than multiple `&&`/`||` with side effects.

eregon:
@pkuczynski Could you take a second look at this fix?

scripts/functions/checksum
```diff
@@ -201,17 +201,27 @@ __rvm_checksum_read()
...
-
     # sha512
     _checksum_sha512="$(   __rvm_db_ "$rvm_path/config/sha512" "$_name" | \command \head -n 1 )"
 
     [[ -n "${_checksum_sha512:-}" ]] ||
       _checksum_sha512="$( __rvm_db_ "$rvm_user_path/sha512"   "$_name" | \command \head -n 1 )"
 
-    [[ -n "${_checksum_sha512:-}" && $_name == http* ]] ||
-       _checksum_sha512="$(__rvm_curl -s -L $_name.sha512)"
+    __rvm_checksum_none || return 0
```
mpapis: this will never get to checking URL's if there was no checksum locally, `__rvm_checksum_none` has to be moved to the end of the function, outside of loops

eregon: Good catch, Bash conditions are hard to read.
```bash
__rvm_checksum_none()
{
  [[ -z "${_checksum_md5:-}" && -z "${_checksum_sha512:-}" ]]
}
```
So that returns "true" if both are unset, i.e, it returns "false" if any is set and therefore `|| return 0` returns if any is set.
This is also done at the function entry BTW.

So that seems incorrect in the case we get one of the checksum, but not the other, right?

I think we should return only if both are set then, i.e., replace that with `__rvm_checksum_all && return 0`. I'll do that.

eregon: @mpapis Should this function return 0 or 1 if we found only one of the checksums?
`__rvm_checksum_validate_file` seems happy if only one is set.

So `__rvm_checksum_any && return 0` would be right (and the original `__rvm_checksum_none || return 0` too) but `_any && return` is probably readable?

eregon: I changed to the equivalent `__rvm_checksum_any && return 0`.

pkuczynski:
Looks better now I think. @mpapis whats your view on this?

eregon:
Here is a build which ended up downloading truffleruby twice because of this bug:
https://travis-ci.org/ruby/mspec/jobs/538473966#L476

eregon:
I'll merge this, it fixes an important bug and allows checksum verification to work again.

pkuczynski:
Sorry for not being able to look at this earlier...

PR #4711, opened 2019-06-11
Use POSIX.2 compatible sed -E for duplicate slash removal

### LINKED ISSUE #4618: Update to latest RVM breaks on OSX
After doing a `rvm get master` RVM no longer activated when I enter any ruby directory, instead I get failures as follows:

```
% cd my-project
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
(eval):1: command not found: cat
(eval):1: command not found: cat
```

The issue appears to be in `scripts/functions/utility` around lines 60-67:
```
  if
    [[ "$PATH" == *"//"* || "$PATH" == *"/:"* ]]
  then
    if [[ "${_system_type}" = "Darwin" ]]
    then PATH="$(\sed -E -e 's#/+#/#g' -e 's#/:#:#' <<<$PATH)"
    else PATH="$(\sed -r -e 's#/+#/#g' -e 's#/:#:#' <<<$PATH)"
    fi
  fi
```

When entering this code `${_system_type}` is blank and `Darwin` is therefore misdetected.

trinitronx:
Fixes #4618 

* Avoid platform detection required for `sed -r` call
* Use POSIX.2 standard flag `sed -E` instead


pkuczynski:
@mpapis what you think?

scripts/functions/utility
```diff
@@ -60,10 +60,7 @@ __rvm_remove_from_path()
   if
     [[ "$PATH" == *"//"* || "$PATH" == *"/:"* ]]
   then
-    if [[ "${_system_type}" == "Darwin" ]]
```
mpapis: the `-E` flag might not be supported on some systems, it might be safer to use here `[[ "$(command uname)" == "Darwin" ]]` - it will be slower, but more reliable

trinitronx: Yeah, it seems like this was why the original bug was happening actually.

According to the discussion in #4618, it was actually the other way around `-r` is not supported on some systems while `-E` should be in most modern versions of `sed` since the POSIX.2 standard.

I think @clive-hetzner summed it up well when he commented:

> I feel that altering the sed behaviour is just (re)hiding the underlying bug that system detection is never executed in this one case - re-entering an rvm folder after having exited it.
> There are other source paths (routes) through the code that could be impacted by the missing system detection in this case.

So the underlying issue he mentioned was that `_system_type` was not set in this case (changing directory to a ruby project).  Therefore the original code assumed `sed -r` was supported and resulted in error:

```
sed: illegal option -- r
```

The other thing we discussed was that the original test is based on platform which is conflated to mean `sed` version.  So the underlying assumption is that `-r` vs. `-E` support is tied to `sed` version & platform.  That assumption turns out to be incorrect, as various versions of macOS have [included various versions of `sed`][1] throughout the years.  Ideally we could detect the version of `sed` or flags supported, but that might be difficult to get right everywhere.

So this proposed solution simplifies things a bit and makes the assumption that POSIX.2 is supported by all the modern *nix platforms.  Not sure if it's the "best" way or if there is one way to support all platforms well without extensive cross-platform testing.

[1]: https://github.com/rvm/rvm/issues/4618#issuecomment-501001444

trinitronx: So to simplify & summarize the problem a bit, there are actually 2 issues:

1. Version / variant of `sed` used across *nix platforms is different, and thus supports different flags.  (POSIX standards could help us out here)
2. Platform detection wasn't being executed properly & could crop up elsewhere

trinitronx: Another alternate idea I just tested to sidestep the issue could be to just use plain old POSIX Basic RegEx (BRE) instead of Extended RegEx (ERE):

```
PATH="$(\sed -e 's/\/\{1,\}/\//g' -e 's/\/:/:/' <<<$PATH)"
```

This should work on all POSIX platforms and even older versions of `sed`, because it's just using the [`"\{m,n\}"` BRE syntax][1] for matching one or more slashes.  It also needs to escape the literal forward slashes `/` due to the default BRE separator being forward slash `/`.  It's a bit more cumbersome to look at which just shows the readability advantages of using [ERE features][2] such as `+` and alternate find/replace separators.  Yet, it's more cross-platform capable using basic POSIX [`sed`][3] features and solves the issue! 😄

@mpapis:  I'd love to hear what you think of this alternate solution, and the other ideas mentioned above.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_06
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04
[3]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

mpapis:
@trinitronx thanks for pointing this out - I was so bent on saving the code from breaking apart in the rare cases I lost the real issue from my eyes - calling to sed to do a shell operation, in 99.99% cases it's faster to run something in shell then to subprocess to other tool like sed or whatever

```bash
# x=a///////b
# echo $x
a///////b
# while [[ "$x" == *"//"* ]] ; do x="${x/\/\///}" ; done ; echo $x
a/b
```

pkuczynski:
@trinitronx can you apply @mpapis suggestion?

pkuczynski:
@trinitronx?

trinitronx:
@pkuczynski: Sure, I'll try to find some time to implement this.

@mpapis: Just to be clear, we're talking about sidestepping the issue using plain Bash substring manipulation instead of a call to `sed`, correct?

trinitronx:
> in 99.99% cases it's faster to run something in shell then to subprocess to other tool like sed or whatever

Looks like it's correct in this case:

```
$ export x="/Users//jcuzella/.rvm/gems/ruby-2.4.0@sprout-wrap/bin/:/Users//jcuzella/.rvm/gems/ruby-2.4.0@global/bin:/Users/jcuzella/.rvm/rubies/ruby-2.4.0/bin:/Users/jcuzella/.rvm/bin/:/usr/local/opt/openssl/bin:/usr/local/sbin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/VMware Fusion.app/Contents/Public:/usr/local/MacGPG2/bin:/opt/X11/bin:/Applications/Wireshark.app/Contents/MacOS"

$ time g

real	0m0.008s
user	0m0.002s
sys	0m0.005s

$ export x="/Users//jcuzella/.rvm/gems/ruby-2.4.0@sprout-wrap/bin/:/Users//jcuzella/.rvm/gems/ruby-2.4.0@global/bin:/Users/jcuzella/.rvm/rubies/ruby-2.4.0/bin:/Users/jcuzella/.rvm/bin/:/usr/local/opt/openssl/bin:/usr/local/sbin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/VMware Fusion.app/Contents/Public:/usr/local/MacGPG2/bin:/opt/X11/bin:/Applications/Wireshark.app/Contents/MacOS"

$ time f

real	0m0.002s
user	0m0.002s
sys	0m0.000s


$ type f
f is a function
f ()
{
    while [[ "$x" == *"//"* ]]; do
        x="${x/\/\///}";
    done;
    while [[ "$x" == *"/:"* ]]; do
        x="${x/\/:/:}";
    done
}

$ type g
g is a function
g ()
{
    x="$(\sed -E -e 's#/+#/#g' -e 's#/:#:#g' <<<$x)"
}
```

trinitronx:
Implemented using @mpapis' [example above][1]

Pushed alternate solution in bda58754a5ae

[1]: https://github.com/rvm/rvm/pull/4711#issuecomment-501798332

trinitronx:
@mpapis, @pkuczynski : Merge conflicts fixed for `CHANGELOG.md`

mpapis:
@trinitronx Thank you, pushed it one step further - dropped the unnecessary `if`. This said - there are no more path operations with subprocesses, if you notice other places that could benefit from this update feel free to open a PR - this is speeding up RVM.

efojs:
@trinitronx I'm on OS X El Capitan, upgraded RVM to 1.29.8 and also stuck with this problem (sed: illegal option -- r). Thank you a lot!

PeterMozesMerl:
I’m coming from  #4711 

Maybe the following will be nothing new but let’s make it sure.

I experienced this issue for the first time only today even though I’ve been using the same RVM, Ruby, Rails, and macOS version for a long time. It was triggered only by a new Rails project.

The issue appears only if the Rails project has either a .ruby-version file or a ruby line in the Gemfile. As these two are amongst the few things I always delete before I add a new Rails project into the repository, my existing Rails project worked fine. 

In case someone is looking for a solution now, just delete these unless you rely on them. I don’t. Moreover, the only issues I’ve ever had with RVM were also related to this file and Gemfile line.

It is common that I use different Ruby versions on my development machines, and another version on the CI and the production servers. (I usually have the same version on the CI and the production servers). On a specific machine, all my projects are running under the same Ruby version though. 

I’m not sure why we have them. I would think it’s more likely that one needs to use a different Ruby version on another platform than that one needs to use a different Ruby version for another Ruby/Rails project.

Maybe they are needed for some commercial hosting service. I don’t know. But as soon as you deploy to your servers, the .ruby-version file is only noise in the VCS at best, and once in a while, a source of issues. 

(I guess they aren’t created by RVM but Rails. I never looked into it. I just delete them for my mental peace.)

PR #4717, opened 2019-06-19
Fix checksums part two: only check URL checksums for Rubinius

eregon:
Now that checksums work again with #4707, I noticed new checksum mismatches, for example
https://travis-ci.org/rvm/rvm/jobs/547675787
Which seems to be the reason the CI currently fails on macOS.

In TravisCI, binary rubies from TravisCI are used on macOS, which meant https://rubies.travis-ci.org/osx/10.13/x86_64/ruby-2.3.4.tar.bz2?rvm=1.29.8-next.md5 would be tried.
And actually, even fixing that to https://rubies.travis-ci.org/osx/10.13/x86_64/ruby-2.3.4.tar.bz2.md5, that still downloads the full archive, not a checksum file.
So let's not assume `"#{archive_url}.md5"` always downloads a checksum file, but only do it for Rubinius which needs it (#4650).
Downloading a checksum from the same location as the archive seems also less good from a security point of view, so limiting that to the only case that needs it is also better.


eregon:
This fixes the CI, I'll merge it.

pkuczynski:
Works for me! Thx!

PR #4728, opened 2019-07-11
Fix rvm-installer's RVM version-fetching from Bitbucket

sfgeorge:
Support for Bitbucket REST API v1 had been dropped.  This fix ensures that if GitHub returns a 403, the Bitbucket fallback can save the day.

3 additional changes are in this PR order to improve the strictness of the fetch_version function, leading to clearer error messages and avoids fetching bogus URLs without tags.

#### Bug fixes
* Prevent attempted download of null RVM versions in rvm-installer.
* Fix rvm-installer's mechanism for fetching RVM tags from Bitbucket.
  _This is increasingly important when occasional 403 / api-rate-limiting issues occur when interacting with GitHub._

#### Changes
* rvm-installer now reports which URL it is fetching version information from.
* rvm-installer now reports when version fetching has completely failed.

### DIFF
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -29,6 +29,7 @@ log()  { printf "%b\n" "$*"; }
 debug(){ [[ ${rvm_debug_flag:-0} -eq 0 ]] || printf "%b\n" "$*" >&2; }
 notice() { log "$*" >&2 ; }
 fail() { log "\nERROR: $*\n" >&2 ; exit 1 ; }
+fail_with_code() { code="$1" ; shift ; log "\nERROR: $*\n" >&2 ; exit "$code" ; }
 
 rvm_install_commands_setup()
 {
@@ -831,7 +832,7 @@ and re-mount partition ${partition} without the noexec option."
 
 rvm_install_select_and_get_version()
 {
-  typeset _version_release
+  typeset dir _version_release _version
 
   for dir in "$rvm_src_path" "$rvm_archives_path"
   do
@@ -842,20 +843,23 @@ rvm_install_select_and_get_version()
   case "${version}" in
     (head)
       _version_release="${branch}"
-      install_head sources[@] ${branch:-master} || exit $?
+      install_head sources[@] ${branch:-master} || fail_with_code $? "Failed to install version '$version'"
       ;;
 
     (latest)
-      install_release sources[@] $(fetch_version sources[@]) || exit $?
+      _version=$(fetch_version sources[@])   || fail "Failed to fetch version '$version'"
+      install_release sources[@] "$_version" || fail_with_code $? "Failed to install version '$version'"
       ;;
 
     (latest-minor)
       version="$(\cat "$rvm_path/VERSION")"
-      install_release sources[@] $(fetch_version sources[@] ${version%.*}) || exit $?
+      _version=$(fetch_version sources[@] ${version%.*}) || fail "Failed to fetch version '$version'"
+      install_release sources[@] "$_version"             || fail_with_code $? "Failed to install version '$version'"
       ;;
 
     (latest-*)
-      install_release sources[@] $(fetch_version sources[@] ${version#latest-}) || exit $?
+      _version=$(fetch_version sources[@] ${version#latest-}) || fail "Failed to fetch version '$version'"
+      install_release sources[@] "$_version"                  || fail_with_code $? "Failed to install version '$version'"
       ;;
 
     (+([[:digit:]]).+([[:digit:]]).+([[:digit:]])) # x.y.z
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -263,7 +263,7 @@ fetch_version()
   return 4
 }
 
-# Returns a sorted list of all version tags from a repository
+# Returns a sorted list of most recent tags from a repository
 fetch_versions()
 {
   typeset _account _domain _repo _url
@@ -272,7 +272,7 @@ fetch_versions()
   _repo=$3
   case ${_domain} in
     (bitbucket.org)
-      _url=https://${_domain}/api/1.0/repositories/${_account}/${_repo}/branches-tags
+      _url="https://api.${_domain}/2.0/repositories/${_account}/${_repo}/refs/tags?sort=-name&pagelen=20"
       ;;
     (github.com)
       _url=https://api.${_domain}/repos/${_account}/${_repo}/tags
@@ -284,7 +284,7 @@ fetch_versions()
   esac
   notice "Fetching ${_url}"
   __rvm_curl -sS ${_url} |
-    \awk -v RS=',' -v FS='"' '$2=="name"{print $4}' |
+    \awk -v RS=',|values":' -v FS='"' '$2=="name"&&$4!="rvm"{print $4}' |
     sort -t. -k 1,1n -k 2,2n -k 3,3n -k 4,4n -k 5,5n
 }
 
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -258,6 +258,8 @@ fetch_version()
       return 0
     fi
   done
+  notice "Exhausted all sources trying to fetch version of RVM!"
+  return 4
 }
 
 # Returns a sorted list of all version tags from a repository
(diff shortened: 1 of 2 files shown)


binscripts/rvm-installer
```diff
@@ -838,20 +843,23 @@ rvm_install_select_and_get_version()
...
       version="$(\cat "$rvm_path/VERSION")"
-      install_release sources[@] $(fetch_version sources[@] ${version%.*}) || exit $?
+      _version=$(fetch_version sources[@] ${version%.*}) || fail "Failed to fetch version '$version'"
+      install_release sources[@] "$_version"             || fail_with_code $? "Failed to install version '$version'"
       ;;
 
     (latest-*)
-      install_release sources[@] $(fetch_version sources[@] ${version#latest-}) || exit $?
+      _version=$(fetch_version sources[@] ${version#latest-}) || fail "Failed to fetch version '$version'"
+      install_release sources[@] "$_version"                  || fail_with_code $? "Failed to install version '$version'"
```
sfgeorge: I broke these `install_release ... $(fetch_version ...)` statements into 2 lines here, so that we can exit (fail_with_code) immediately after returning from `fetch_version`.

Alternatively, I could have just `exit`ed directly within `fetch_version` on line 263 when it detects failure.  But I am always hesitant to add `exit` to additional functions, as it adds complexity.

Still, I'm open to changing this to exit early in fetch_version if it helps reduce the delta.

binscripts/rvm-installer
```diff
@@ -257,9 +259,11 @@ fetch_version()
       return 0
     fi
   done
+  notice "Exhausted all sources trying to fetch version of RVM!"
```
pkuczynski: This should rather use `fail_with_code`, right?

binscripts/rvm-installer
```diff
@@ -278,8 +282,9 @@ fetch_versions()
       _url=https://${_domain}/api/v3/repos/${_account}/${_repo}/tags
       ;;
   esac
+  notice "Fetching ${_url}"
```
pkuczynski: Why just not `log`?

sfgeorge: Good question there. I discovered actually necessary to avoid emitting to the STDOUT stream in this case, because `fetch_version` captures the STDOUT output of `fetch_versions`, and assigns it to `_version`.

binscripts/rvm-installer
sfgeorge: That's a fair Q as well, and relates directly to https://github.com/rvm/rvm/pull/4728#discussion_r302506596. I can definitely simplify some logic by just using `fail_with_code` here and be over & done with it.  I was just hesitant to add an implicit `exit` to a function where it hadn't been there yet before.

But there are Pros for either approach.  What's your pref?  I am happy to lean either way.

pkuczynski: @mpapis what you think?

binscripts/rvm-installer
pkuczynski: Maybe use debug then? @mpapis?

binscripts/rvm-installer
pkuczynski: `notice` and `log` feels quite similar just by looking at the name, so I would not bring `notice` on board

sfgeorge: I was thinking it would be helpful to log problems to STDERR instead of STDOUT.  The thought being to use the same output stream as `fail`, but without halting.

binscripts/rvm-installer
sfgeorge: With debug, regular users / observers in Travis will not see the message.  My intent was to make sure that they can see what is being fetched, as it avoids to obscure 403 Forbidden and 410 Gone errors like in the [failed build](https://travis-ci.org/adhearsion/adhearsion/jobs/555932810) below:

failed travis build

I think having the extra noise may help people more quickly figure out what's going on when API calls may fail.

I otherwise would not have known through a Travis build that the 403 Forbidden was from the GitHub API and the 410 Gone was from the Bitbucket v1 API.

binscripts/rvm-installer
```diff
@@ -268,7 +272,7 @@ fetch_versions()
   _repo=$3
   case ${_domain} in
     (bitbucket.org)
-      _url=https://${_domain}/api/1.0/repositories/${_account}/${_repo}/branches-tags
+      _url="https://api.${_domain}/2.0/repositories/${_account}/${_repo}/refs/tags?sort=-name&pagelen=20"
```
sfgeorge: FYI I added `&pagelen=20` to increase from the default page size of 10.  We could go as high as fetching the 100, but I thought that might be excessive (unless someone really wants to install a version from long ago).  But it any case, it is not possible to go and fetch _all_ tags in one go, they would have to be accumulated over multiple requests.



> pagelen: Current number of objects on the existing page. Globally, the minimum length is 10 and the maximum is 100. Some APIs may specify a different default.

https://developer.atlassian.com/bitbucket/api/2/reference/meta/pagination

binscripts/rvm-installer
pkuczynski: I agree that producing valid error output is crucial to find problems easily. I am just not such a big fan of introducing many different logging functions, as one gets lost quickly which one should be used...

binscripts/rvm-installer
pkuczynski: I would let @mpapis decide on this.

But if we keep it, we should rather call the function `error` not `notice`, don't you think?

binscripts/rvm-installer
pkuczynski: I believe 20 is enough as it goes a few years back :) We do not release too often.

binscripts/rvm-installer
sfgeorge: Good call.  Though only trouble with `error` is that my change on line 285 becomes especially awkward, when in fact in that case we don't yet _know_ whether the request is going to fail or succeed:

    error "Fetching ${_url}"

binscripts/rvm-installer
sfgeorge: If it helps, I can attempt to retroactively print the URL, after curl has emitted a failure.  It would look a bit more like this if I can pull it off:

> curl: (22) The requested URL returned error: 403 Forbidden
> **... Fetching https://api.github.com/repos/rvm/rvm/tags has failed.**
> curl: (22) The requested URL returned error: 410 Gone
> **... Fetching https://api.bitbucket.org/2.0/repositories/mpapis/rvm/refs/tags?sort=-name&pagelen=20 has failed.**
> Downloading https://github.com/rvm/rvm/archive/.tar.gz

Then I could make use of STDOUT / `log` to print these.  Extra noise would only be printed when a failure or recoverable failure has occurred.

Should I give that a stab?

pkuczynski: Ideally, it should print `Fetching https://api.github.com/repos/rvm/rvm/tags has failed (403 Forbidden)` when curl failed, so basically `Fetching {url} ({error status curl})`

sfgeorge: I can give that a go.   Though I'm a little nervous about printing "Fetching .." prior to the curl error output, because it requires that I first capture the curl error output off of STDERR to a variable, then print "Fetching .." along with the variable's contents.  (If any part of that process fails, then we might halt before we get to print the original curl error).

I'll give it a try as a separate commit, orphaned out of the this branch / PR.  If we like how it looks, then we can include it in this PR (that way there's fewer unnecessary travis builds).

One additional Q: Should we provide the URL upon failure only in `fetch_versions`, or for all calls to `__rvm_curl`?  I think I can handle either one.  Though if we go with the latter, we may want a more generic message like:
> Request {url} has failed ({error status curl})

For now, I'm doing to draft it up as modification only to `fetch_versions` as we discussed, but I can apply it more generically to `__rvm_curl` if we like that idea.

sfgeorge: head's up: I attempted this and the situation [got pretty complicated](https://github.com/sfgeorge/rvm/compare/fix-rvm-installer-fetch-version-bitbucket...fix-rvm-installer-fetch-version-bitbucket.selective-noise) - I have yet to come up with a clean solution to honoring this without using STDERR, since STDOUT is reserved for emitting the version from `fetch_version`.

Since this is a difficult topic that requires more consideration, I am going to close this PR and split it into 2 distinct purposes 1.) Fix Bitbucket (urgent due to many GitHub 403 rate-limiting errors on Travis) and 2.) Improve error detection & messaging.

PR #4731, opened 2019-07-11
Improve rvm-installer error handling & error messaging

sfgeorge:
3 changes in order to improve the strictness of the fetch_version function, leading to clearer error messages and avoids fetching bogus URLs without tags.

#### Bug fixes
* Prevent attempted download of null RVM versions in rvm-installer.

#### Changes
* rvm-installer now reports which URL it is fetching version information from.
* rvm-installer now reports when version fetching has completely failed.


pkuczynski:
@sfgeorge could you please solve conflicts?

binscripts/rvm-installer
```diff
@@ -27,7 +27,9 @@ rvm_install_initialize()
 
 log()  { printf "%b\n" "$*"; }
 debug(){ [[ ${rvm_debug_flag:-0} -eq 0 ]] || printf "%b\n" "$*" >&2; }
+notice() { log "$*" >&2 ; }
 fail() { log "\nERROR: $*\n" >&2 ; exit 1 ; }
```
pkuczynski: This should probably change to `fail_with_code 1 "$*"`, right?

binscripts/rvm-installer
```diff
@@ -27,7 +27,9 @@ rvm_install_initialize()
 
 log()  { printf "%b\n" "$*"; }
 debug(){ [[ ${rvm_debug_flag:-0} -eq 0 ]] || printf "%b\n" "$*" >&2; }
+notice() { log "$*" >&2 ; }
```
pkuczynski: If we writing to STDERR, maybe we should call this function `error()`?

binscripts/rvm-installer
```diff
@@ -278,7 +281,8 @@ fetch_versions()
       _url=https://${_domain}/api/v3/repos/${_account}/${_repo}/tags
       ;;
   esac
-  __rvm_curl -sS ${_url} |
+  notice "Fetching ${_url}"
```
mpapis: Maybe `debug`? It's hard to find balance between enough and to much information, dbug help with it as you can add flag to display the extra messages.

pkuczynski: Hmm, I am not sure if it makes sense to push that to STDERR?

sfgeorge: Since `rvm_install_select_and_get_version` captures all STDOUT from `$(fetch_version ...)` and interprets all of that output as a version specifier, we cannot emit warning information about GitHub failures to STDOUT; it will taint the results of the version that gets successfully retrieved from the BitBucket fallback.

Therefore neither `fetch_version` nor the `fetch_versions` function that it calls can emit warning information to STDOUT; they would have to emit to STDERR.

> If we writing to STDERR, maybe we should call this function error()?
 Can I call it `warn()`? The characteristics needed for this function are:
* It writes to STDERR.
* It does not exit.
 
Feels like `warn()` may be a good fit.  What do you guys think?

sfgeorge: FYI In response to https://github.com/rvm/rvm/pull/4728#discussion_r302555052 , I'm in the midst of replacing this "Fetching ..." log message (which would be noisy for completely successful runs :-1:) with a message that only emits when a [recoverable] fetch_versions failure has occurred. _(PR to soon have some additional commits for review ✨ )_

Given that the message will only be emitted after a [recoverable] failure has occurred, I think that debug would be too silent. I think that we'd want all users to know about the [recoverable] failure, even under normal verbosity.

binscripts/rvm-installer
sfgeorge: Good call sir, fixed

binscripts/rvm-installer
```diff
@@ -278,7 +281,8 @@ fetch_versions()
       _url=https://${_domain}/api/v3/repos/${_account}/${_repo}/tags
       ;;
   esac
-  __rvm_curl -sS ${_url} |
+
+  { __rvm_curl -sS "${_url}" || warn "...the preceeding error with code $? occurred while fetching $_url" ; } |
```
sfgeorge: @pkuczynski I've updated this message to appear like so when a GitHub fetch_versions() fails but a BitBucket succeeds:

screenshot-of-recoverable-fetch-failure

It has the following characteristics:
* Only emits log output when something bad (recoverable) happens.
* Emits to STDERR, to avoid breaking the STDOUT capture of _version=$(fetch_version ...)

I believe this honors spirit of what we discussed in https://github.com/rvm/rvm/pull/4728#discussion_r302555052

However, it deviates slightly from your original suggestion, which involved deferring the display of the curl STDERR ({error status curl}) output until after most of the log message:
> Ideally, it should print Fetching https://api.github.com/repos/rvm/rvm/tags has failed (403 Forbidden) when curl failed, so basically Fetching {url} ({error status curl})

----

If desired, I do have [this alternative version](https://github.com/sfgeorge/rvm/compare/improve-rvm-installer-error-handling...sfgeorge:improve-rvm-installer-error-handling.defer-stderr) which does honor your request and emits the curl STDERR message last, by first capturing it and then displaying it.  However, it proves quite difficult to capture both STDOUT and STDERR separately in bash without mushing them together into one stream.  [Most suggests out there](https://stackoverflow.com/questions/11027679/capture-stdout-and-stderr-into-different-variables) involve some use of eval().  More reasonable suggestions involve capturing to named-pipes, or to simple files.  My alternate approach opts for the latter for simplicity - dumping the STDERR output to a file, and emitting that if a non-zero exit code is detected.  I'm keeping this approach out of the PR for now since it has a few potential draw-backs:
* Adding an additional file to write to the user's disk seems less than ideal.
* If the capturing & outputting mechanism I've added fails for whatever reason (such as disk full or permission denied), then curl's STDERR will have been captured but never displayed.

Let me know your thoughts on a preferred approach.

binscripts/rvm-installer
sfgeorge: I've added a `warn()`, let me know what you think.  I can rename to `error()` if you like.

binscripts/rvm-installer
sfgeorge: ...updated.  Let me know what you think!

pkuczynski: I think STDERR is meant to include errors and not warnings? @mpapis what would u suggest to do in this case?

binscripts/rvm-installer
pkuczynski: I think it's not worth to go so far as in the alternative PR. What you did here should be good enough - even if it deviates from the nicest output it should be helpful enough I think.

pkuczynski:
@sfgeorge thanks for all your hard work on this! Maybe you would like to become a maintainer?

binscripts/rvm-installer
sfgeorge: @pkuczynski Thanks very much, man.  I appreciate it!

On an unrelated note, I think the [merge into master unfortunately broke](https://travis-ci.org/rvm/rvm/builds/558900224) the signature for rvm-installer; and someone has already noticed (#4736).  Would you mind re-signing to make it all happy?

sfgeorge:
@pkuczynski Yes, that would be awesome!  I'd love to get more involved and help out.  Thank you!! =)

binscripts/rvm-installer
sfgeorge: @pkuczynski or @mpapis Can you update the signature for rvm-installer.asc?  This is unfortunately affecting some builds (#4736).

pkuczynski:
@sfgeorge you should get an invite. What's your email so I can invite you for our slack?

sfgeorge:
@pkuczynski Thank you!  Let me know if this is too much of a pain, I can share via an alternate method.

```
-----BEGIN PGP MESSAGE-----

hQIMA1/IIr9lIf8XAQ//YQvM56WqxwqL6WjLoFECyABJHK8daVWlC8jsLYUfYEBs
LTfRxw+XMbG8AeaZoW+rhXEovj2gPP5oA2X5VE1xsTdwDXVHkUU0la12lhaqh95c
UxnY/5CmtriLwE4VA7Lq9OjKCA/j93TPSWUvb+Z0zWZqGkMbf66swQKHOzXHc5NS
7XSl3muCsxd+qoEZtYv3DumTRH6CJnCeuUOPEjtBph12PnLQifd8arH3fxr4wyDW
WEnEiwGcAUC1mnRNCxffw2Mj2QT9tpw+Id9HuLTTaXesJI4FydD018MtnPwW+JMm
GvaJZWGmvhIQHIvpI7n7nU7lso+WxdSCaC4C3ydf0dtn63pFLMqnnC1A2ny31VTQ
lTgPZY/jBdetILvfJU9H9bx7YQw314ficIZ5XzSSymxxYU9E4omPiC3zTAnmROs/
+bzkZxQCWoLuYxZsGtbTiQYl+Ni9zALmNMaR4AgzYO4nYciMDZTTRuTy+pT7bMH/
YkN36uOvxugQ8h5Gh2YQvWwogvXzoRcbXzLm8Nkq0NdGs+Yvj4T4tHjcsevnrXB5
9Sp8Gn6EApTc+WUP1T41eYIYhpSqNMznlIWKjZCsBdNW5s7o8I7grbmLlyNdQhug
5ZYj5D8Qa02T9RZ5Mgc/TQp0xHnRDGzzTi6U5TPpfyKrfjYI2pqaWK1nG0BTmoLS
YgEZ4lQ8Z/yKlXxz9VvHQMYMP0nj0yoy40R/Us2BfCddLM+MggnXw6yuL2DeZh2r
qLXK0pxSGEh/7dOHPLZxvccblD9cqm9dAAPDT0o0UFgd7jgdNv70Nh4Br3VmIf1G
PMLq
=Jmzo
-----END PGP MESSAGE-----
```

pkuczynski:
No pain, no gain ;) Invite sent :)

PR #4733, opened 2019-07-14
Use doas for openbsd.

### LINKED ISSUE #4732: OpenBSD attemtps to use sudo instead of doas, and fails.
## Description

When running rvm, we get errors about using sudo:
```
Installing requirements for openbsd.
Updating system....Running 'tar -C /usr -zxf /tmp/tmp.WdgFXoPKG66OPgbtkt' would require sudo, but 'sudo' is not found!
..
Error running 'requirements_openbsd_update_system ruby-2.5.3',
please read /home/ori/.rvm/log/1563077206_ruby-2.5.3/update_system.log
Requirements installation failed with status: 1.
```

## Steps to reproduce

I'm running on a current OpenBSD snapshot, but this should reproduce anywhere doas is used in preference to sudo:

Run any command that requires sudo from rvm.

## Expected behavior

I expect doas to silently be used in place of sudo.

## Actual behavior

RVM tries to find sudo, fails, and errors.

oridb:
It's the new hotness, we no longer have sudo shipped by default on openbsd. The doas command first appeared in openbsd 5.8.

Fixes #4732 .

* Add a probe for `doas`, use it as a fallback to sudo. Default to sudo in rvm.sh on OpenBSD.


scripts/functions/support
```diff
@@ -356,6 +356,19 @@ __rvm_setup_sudo_function_Other()
     {
       \command \sudo "$@"
     }
+  elif
```
pkuczynski: Instead of modifying `__rvm_setup_sudo_function_Other` you should add new function

```
__rvm_setup_sudo_function_OpenBSD()
{
...
}
```

and implement this logic only there.

binscripts/rvm-installer
```diff
@@ -46,7 +46,7 @@ rvm_install_commands_setup()
       (OpenBSD)
         log "Trying to install GNU version of tar, might require sudo password"
         if (( UID ))
-        then sudo pkg_add -z gtar-1
+        then doas pkg_add -z gtar-1
```
pkuczynski: We should add some `does` detection logic here too:

```
if (( UID ))
then 
  if \which doas >/dev/null 2>&1
    then doas pkg_add -z gtar-1
    else sudo pkg_add -z gtar-1
  fi
else
  pkg_add -z gtar-1
fi
```

scripts/functions/support
oridb: I thought about that, but doas is available on other platforms (eg, https://www.freebsd.org/cgi/man.cgi?query=doas&sektion=1&manpath=freebsd-release-ports). It seems like if it's there, we should use it.

oridb:
ping?

(The test failures look odd to me, and I'm not sure how this change could have broken them.)

scripts/functions/support
pkuczynski: I don't think this is solved the way I suggested?

pkuczynski:
@oridb sorry for delay. I made one more comment, as I don't think you solved the previous one the way I suggested? Do you need help with this?

morgant:
@oridb Any update on changing the implementation as @pkuczynski suggested in his initial code comment? I'd love to test rvm on OpenBSD.

schrodyn:
I'm hitting this issue but with doas on FreeBSD. Can you include FreeBSD + doas support too, please.

PR #4740, opened 2019-07-22
Update Truffleruby to 19.1.1

skunkworker:
Updated Truffleruby to 19.1.1.

Unfortunately it looks like new releases 19.1.1+ don't include `truffleruby` in the .tar.gz name anymore but are like `ruby-standalone-svm-linux-amd64-19.1.1.tar.gz`.


pkuczynski:
@eregon?

eregon:
There was no change for TruffleRuby in 19.1.1 vs 19.1.0, so I'd rather not add it to any Ruby manager.
Is there a particular reason you want to use 19.1.1 instead of 19.1.0?

You are right, the filename is incorrect, we'll fix it.

pkuczynski:
I will close it for now, but you can continue discussing details here...

ezzarghili:
The release artifacts names for the standalone versions were updated to follow the consistent name used before, sorry for the inconvenience.
```
truffleruby-19.1.1-macos-amd64.tar.gz
truffleruby-19.1.1-linux-amd64.tar.gz
```

Issue #4282, opened 2018-01-16
`rvm install <version>` fails to regenerate gem wrappers when `ls` has been aliased in certain ways


rdodson41:
## Description

`rvm install <version>` fails to regenerate gem wrappers when `ls` has been aliased in certain ways.

## Steps to reproduce

 1. Alias `ls` as follows:

```
$ alias ls='"/usr/local/opt/coreutils/libexec/gnubin/ls" --format=long --human-readable --time-style=long-iso --classify --color=auto'
```

 2. Run `rvm install ruby-2.4.3`.

## Expected behavior

* [`rvm install ruby-2.4.3`](https://gist.github.com/rdodson41/53e0a382e69045b55ee17048c9770f28)

Note the relevant successful steps:

```
ruby-2.4.3 - #importing gemset /Users/rich/.rvm/gemsets/global.gems.............................................
ruby-2.4.3 - #generating global wrappers.......
```

## Actual behavior

* [`rvm install ruby-2.4.3`](https://gist.github.com/rdodson41/d62f413b0f8220ce5e79b28864a5a8c9)

Note the relevant errors when `rvm` attempts to regenerate gem wrappers:

```
ruby-2.4.3 - #importing gemset /Users/rich/.rvm/gemsets/global.gems.............there was an error installing gem gem-wrappers
.................................there was an error installing gem rvm
.
ruby-2.4.3 - #generating global wrappers.............
Error running 'run_gem_wrappers regenerate',
please read /Users/rich/.rvm/log/1516069374_ruby-2.4.3/gemset.wrappers.global.log
```

* [`gemset.wrappers.global.log`](https://gist.github.com/rdodson41/a81a9ce6f012488ea3a2d2a03e3c7a28)

Note the relevant errors:

```
++ printf %b 'command(7): command gem install -rw-r--r-- 1 rich staff 13K 2018-01-15 21:20 /Users/rich/.rvm/gem-cache/gem-wrappers-1.4.0.gem --local --no-ri --no-rdoc\n'
++ rvm_log 'installing gem -rw-r--r-- 1 rich staff 13K 2018-01-15 21:20 /Users/rich/.rvm/gem-cache/gem-wrappers-1.4.0.gem --local --no-ri --no-rdoc'
++ [[ '' == 1 ]]
[...]

PR #4743, opened 2019-07-24
Ignore aliases when using `ls` command

skunkworker:
This is an addendum on #4282 for ZSH and possibly other shells.
The current version works fine on bash when `alias ls='something-not-ls'` but on zsh `alias ls='something-not-ls'` is called it currently fails.

It also fixes another `ls` command invocation in `is_gem_installed()`. 

Where `/bin/ls` is installed, this is used instead of the terminals `ls` short command. Otherwise it defaults to the shell sessions `ls` alias if it doesn't find it.

This can be helpful to make sure the default command doesn't have extra arguments.

This was my first attempt and I was able to get rid of the bug, but I think it might be worth it to move the `_ls_command` command out of the function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix installer fetching RVM tags from Bitbucket [\#4730](https://github.com/rvm/rvm/pull/4730)
 * Prevent downloading of null RVM versions in installer [\#4731](https://github.com/rvm/rvm/pull/4731)
 * RVM package fails to configure on fresh Ubuntu 18.04 server install [\#4742](https://github.com/rvm/rvm/pull/4742)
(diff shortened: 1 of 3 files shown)


skunkworker:
> We write wrappers over system commands in `scripts/functions/support`. In this the best case would be to define it in `__rvm_setup_utils_functions_common()`:
> 
> ```
> __rvm_ls() { \command \ls "$@" || return $?; }
> ```
> 
> This definition should ignore aliases. Then you should be able to use `_rvm_ls` in `functions/gemset`. Can you give it a try please?

Thanks @pkuczynski this seemed to work on my ZSH environment just fine with the aliasing.

pkuczynski:
Looks better! @mpapis what you think?

We are only missing CHANGELOG entry...

PR #4756, opened 2019-08-14
 Add support for JRuby 9.2.8.0

koic:
JRuby 9.2.8.0 has been released.
https://www.jruby.org/2019/08/12/jruby-9-2-8-0.html

And follow up #4664 💦 The commit ff09091 fixes a sha512 definition for JRuby 9.2.7.0.
This is not the definition of Ruby 2.4.7.


pkuczynski:
Thanks!

PR #4761, opened 2019-08-21
Add TruffleRuby 19.2.0

eregon:
Add TruffleRuby 19.2.0


eregon:
(I made a PR to check if the CI passes)

pkuczynski:
Build fails, so I am not merging just yet...

eregon:
I'd say let's merge since the failure is only on macOS and seems completely unrelated. Is it a known issue though?

eregon:
There seems to be the same failure on master: https://travis-ci.org/rvm/rvm/builds/574280336
@mpapis Maybe we should just `allow_failures` that job until it's fixed?

PR #4766, opened 2019-08-24
Add tests for ruby -S bundle for TruffleRuby and do not install rubygems-bundler by default

### LINKED ISSUE #4765: RVM breaks TruffleRuby launchers and `ruby -S bundle`
This is quite similar to #4408 (which has more details) and has been reported in https://github.com/oracle/truffleruby/issues/1697.
RVM breaks usages of `ruby -S bundle` and notably `lib/cucumber/rake/task.rb`.

The problem is RVM changes shebangs which break e.g. `bundle` on TruffleRuby, because it changes the shebang line from
```
#!/usr/bin/env bash
...
Bash code
```
to
```
#!/usr/bin/env truffleruby_executable_hooks
...
Bash code
```

And this looks to a ruby interpreter like the following code would be Ruby code and not Bash, because there is 'ruby' in the shebang line.

@mpapis Could you explain why is the shebang changed?
What part/gem of RVM does this, could we just skip it for TruffleRuby?
What functionality would we lose if we do so?

eregon:
Fixes #4765

* Add tests for `ruby -S bundle` for TruffleRuby (failing originally, see [this run](https://travis-ci.org/eregon/rvm/builds/576230778))
* Do not install rubygems-bundler for TruffleRuby
* Add a post-install message about rubygems-bundler

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ matrix:
       script:
         - ruby -v
         - rake --version
+        - ruby -S bundle --version
         - ruby -ropen-uri -e 'puts open("https://rubygems.org/") { |f| f.read(1024) }'
 notifications:
   email:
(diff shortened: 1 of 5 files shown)


eregon:
`Test TruffleRuby in TravisCI` fails but that's expected because it uses RVM `master` and this PR is not yet in master. (it's circular unfortunately, will be fixed once this PR is merged)
`truffleruby_comment_test.sh` will now pass on Linux and macOS, showing this solves the issue.

PR #4769, opened 2019-08-28
Fixes checksums for ruby 2.6.4

pcarn:
#4768 added Ruby 2.6.4, but used the checksums from the .tar.gz package instead of .tar.bz2, so it's failing to install.

@koic note that .tar.bz2 should be used.

@pkuczynski Ruby 2.6.4 fails to install until this is merged.

```
ruby-2.6.4 - #downloading ruby-2.6.4, this may take a while depending on your connection...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13.7M  100 13.7M    0     0  24.2M      0 --:--:-- --:--:-- --:--:-- 24.2M
Downloaded archive checksum did not match, archive was removed!
If you wish to continue with not matching download add '--verify-downloads 2' after the command.

There has been an error fetching the ruby interpreter. Halting the installation.
```


koic:
@pcarn That's my fault 💦 I should have used tar.bz2 instead of tar.gz. Thank you for the patch.

PR #4780, opened 2019-09-12
Patch the 2.6.4 string corruption bug

havenwood:
This patches the Ruby 2.6.4 string corruption bug by default for Ruby 2.6.4.

https://bugs.ruby-lang.org/issues/16136


pkuczynski:
@havenwood this patch is exactly the same as provided by @skaes already in #4778. Compare with https://raw.githubusercontent.com/skaes/rvm-patchsets/master/patches/ruby/2.6.4/railsexpress/05-use-after-free-bug-in-symbol-c.patch

havenwood:
Yes, it's the same patch as the railsexpress patch in the gist, except here it's shipped with RVM and installed by default with 2.6.4. I thought it better to install a default patch from local source rather than a gist.

PR #4794, opened 2019-10-01
Bump the default Ruby version to 2.6.5

havenwood:


pkuczynski:
Changelog?

havenwood:
I saw this as part of the last log entry for adding 2.6.5. Should it have a separate entry?

pkuczynski:
Hmm., ok, will merge it

PR #4807, opened 2019-10-28
Run all JRuby commands from gemset down using --dev flag.  This reduces

enebo:
Run all JRuby commands from gemset down using --dev flag. This reduces gemset install
time substantially (50.9s -> 40.5s locally).  The rationale
for this change is that short-running JRuby processes should use --dev because
it disables a bunch of optimizations and essentially just runs using the
startup interpreter.  This hurts eventual performance but it also speeds up
short-lived processes.

I do not think I did this quite right but I put it in as a starting point.
mpapis helped me find the proper place but I need guidance on proper code
itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 [Full Changelog](https://github.com/rvm/rvm/compare/1.29.9...HEAD)
 
 #### New features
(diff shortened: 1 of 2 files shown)


enebo:
Doh!  I did not see the CHANGELOG or other boiler plate.  I will update once we hammer out what needs to change.

PR #4823, opened 2019-12-03
Ruby 2.3.8 should be able to use openssl 1.0

whs-dot-hk:
Fixes # 4781.

* Force openssl to use 1.0.2t
* Instead of 1.1
*

### DIFF
--- a/scripts/functions/requirements/osx_brew
+++ b/scripts/functions/requirements/osx_brew
@@ -307,7 +307,7 @@ requirements_osx_brew_define_openssl()
 
   case "$1" in
     (ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
-      brew_openssl_package="openssl"
+      brew_openssl_package="https://raw.githubusercontent.com/Homebrew/homebrew-core/8c9b113bc6812dc74c598c8f860017e42fba8d78/Formula/openssl.rb"
       ;;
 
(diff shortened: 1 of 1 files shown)


scripts/functions/requirements/osx_brew
```diff
@@ -307,7 +307,7 @@ requirements_osx_brew_define_openssl()
 
   case "$1" in
     (ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
-      brew_openssl_package="openssl"
+      brew_openssl_package="https://raw.githubusercontent.com/Homebrew/homebrew-core/8c9b113bc6812dc74c598c8f860017e42fba8d78/Formula/openssl.rb"
```
pkuczynski: This could maybe work but should be rather done behind some IF, as someone might have older macOS where this openssl version is simply available.

Also check https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula/4158763#4158763

whs-dot-hk: Hi, I should add a openssl version check

but more problem came alone as other homebrew package e.g. mariadb use openssl 1.1

`/usr/local/opt/openssl` will then be link to 1.1

so again ruby 2.3 won't work

c9s: yes, I tried this and it can not be linked correctly.

PR #4837, opened 2019-12-25
Add support for Ruby 2.7.0

koic:
Ruby 2.7.0 has been released.
https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/


koic:
Thank you!

PR #4871, opened 2020-02-15
Add support for truffleruby-head

eregon:
* Uses the latest nightly build, which is automatically built.
* I checked in Docker and it works well with `rvm install truffleruby-head`.

cc @gogainda @deepj

Ref: https://github.com/oracle/truffleruby/issues/1483 https://github.com/oracle/truffleruby/issues/1903


eregon:
CI is green, except for the existing failures on macOS, same as `master`.

PR #4910, opened 2020-04-14
Add support for truffleruby 19.0.2, 19.1.1, 19.2.1, 19.3.0.2

pkuczynski:


config/known
```diff
@@ -33,6 +33,7 @@ rbx-4[.12]
 rbx-head
 
 # TruffleRuby
+truffleruby-19[.3.1]
```
eregon, line 36: I wouldn't add this, anyone using TruffleRuby should use the latest release

eregon:
Looks good. Note that most of these releases are Critical Patch Updates releases, i.e., 19.x.y is the same as 19.x.0 + some security fixes.

In practice, TruffleRuby is still experimental and so using 19.x.y doesn't make much sense, 20.0.0 has the security fixes and is far more compatible.
That's why I didn't bother to add those releases, but we can add them, that's fine too.
So currently, we consider only the latest release of TruffleRuby to be the one people should use, previous versions have simply more bugs/are less compatible and we don't maintain them beyond security fixes.

19.3.0.2 is 19.3.0 + some bug fixes which don't seem particularly relevant for TruffleRuby: https://www.graalvm.org/docs/release-notes/19_3/

See https://www.graalvm.org/docs/release-notes/version-roadmap/ for details.

pkuczynski:
I prefer to keep all hashes for security reasons.

PR #4976, opened 2020-09-03
Add few built-in command aliases to improve consistency

AndyObtiva:
* Alias `rvm gemset remove` as `rvm gemset delete`
* Alias `rvm gemset move` as `rvm gemset rename`
* Alias `rvm delete` as `rvm remove`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * Added railsexpress patches for Ruby 2.5.8, 2.6.6 and 2.7.1 [\#4900](https://github.com/rvm/rvm/pull/4900)
 * Add ruby-3 to the list of available binary builds [\#4984](https://github.com/rvm/rvm/pull/4984)
 * Recognize `3*` as CRuby version number [\#4987](https://github.com/rvm/rvm/pull/4987)
+* Alias `rvm gemset remove` as `rvm gemset delete`
+* Alias `rvm gemset move` as `rvm gemset rename` 
+* Alias `rvm delete` as `rvm remove`
 
 #### Bug fixes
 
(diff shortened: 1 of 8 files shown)


pkuczynski:
Why would you like to propose these changes? They would only make rvm more complex to use...

AndyObtiva:
I hope you’re not serious as I have been using those changes everyday since I contributed the Pull Request. It was more complex to remember when to use delete and when to use remove. Now, I always use delete in both cases of removing a Ruby or a Gemset, so I never have to worry about remembering. As such, they made RVM easier to use. And, if some people reflexively prefer remove, them that works too, which is in line with the Ruby philosophy of more ways to accomplish the tasks to match people’s preferences just like Array#count, Array#length, and Array#size for example. I am not sure if you are a true Rubyist as I saw on your profile you are a fan of node and react, which true Rubyists absolutely despise. Anyways, take my changes or leave them, but do not accuse them of making RVM more complex because they clearly don’t and it would have been obvious to a true Rubyist.

pkuczynski:
I am not sure why anybody would despise some other technologies only because he is not using them. Ruby and React in this example are completely incomparable.

@havenwood @mpapis what do you think about proposed change?

mpapis:
I'm fine with it, we already have few aliases like this for example rm and remove, adding delete might make it more familiar for users of other tools.

pkuczynski:
@AndyObtiva could u please fix CHANGELOG conflict before we can merge?

AndyObtiva:
OK, I just fixed the CHANGELOG.md

pkuczynski (approved):

pkuczynski:
Thanks!

PR #4995, opened 2020-10-24
Update brew list command to remove deprecation warning

### LINKED ISSUE #5007: Warning: Calling `brew list` to only list formulae is deprecated! Use `brew list --formula` instead.
Homebrew has deprecated the old CLI convention. Please make sure to supply the `--formula` flag where appropriate.

miketheman:
Homebrew has begun emitting a deprecation warning when checking for
requirements on osx as of Homebrew/brew#8851

Example output when installing a Ruby:

```
rvm install 2.7.2
Searching for binary rubies, this might take some time.
No binary rubies available for: osx/10.15/x86_64/ruby-2.7.2.
Checking requirements for osx.
Warning: Calling `brew list` to only list formulae is deprecated! Use `brew list --formula` instead.
... (repeated ~12 times on my system)
Warning: Calling `brew list` to only list formulae is deprecated! Use `brew list --formula` instead.
Certificates bundle '/usr/local/etc/openssl@1.1/cert.pem' is already up to date.
Requirements installation successful.
...
```

This does not impact the execution, and the command completes
successfully, however it's extra noise for the end user.

The `--formula` flag was added back in August in Homebrew/brew#8229
and brew will auto-update in most scenarios - but users who have set the
environment variable to not auto-update brew may fail this command with:

    Error: ambiguous option: --formula

I do not have any insight when brew may handle the deprecation warning
and remove the existing behavior.

Fixes #5007


miketheman:
I've added an entry to the CHANGELOG.md - but am uncertain how to interpret the other messages from the Mergable bot - so would appreciate any pointers you may have.

havenwood:
> I've added an entry to the CHANGELOG.md - but am uncertain how to interpret the other messages from the Mergable bot - so would appreciate any pointers you may have.
I added a Label and Milestone for Mergeable.

Any idea how old the `--formula` flag is? Will this be backward compatible to still work with very old versions of Homebrew? If so, LGTM.

Thanks for the PR!

miketheman:
> Any idea how old the `--formula` flag is? Will this be backward compatible to still work with very old versions of Homebrew?

Thanks - I called that out in the PR description:

> The --formula flag was added back in August in Homebrew/brew#8229 ...

felixbuenemann:
The brew commit that added the deprecation warning is https://github.com/Homebrew/brew/commit/c8809ce730ed02f79c0efff4b76ba5e5207e5432.

I don't think it makes sense, to document something as the default behavior and then throw a deprecation warning if you actually do depend on that default, which I've mentioned to @Akylzhan in a comment to the commit.

Added to that this breaks backwards compatibility, which makes the change even worse.

pkuczynski:
> I've added an entry to the CHANGELOG.md - but am uncertain how to interpret the other messages from the Mergable bot - so would appreciate any pointers you may have.

Mergeable fixed

pkuczynski:
> > Any idea how old the `--formula` flag is? Will this be backward compatible to still work with very old versions of Homebrew?
> 
> Thanks - I called that out in the PR description:
> 
> > The --formula flag was added back in August in [Homebrew/brew#8229](https://github.com/Homebrew/brew/pull/8229) ...

Maybe we should check `brew --version` and only add the flag for older homebrew versions? Can you add this please?

miketheman:
> Maybe we should check `brew --version` and only add the flag for older homebrew versions? Can you add this please?

I took a look at this, and added a detection mechanism. I'm not certain of the idioms used in the codebase, but this seemed to match similar cases of version comparison.

scripts/functions/requirements/osx_brew
```diff
@@ -39,7 +39,13 @@ requirements_osx_brew_lib_installed_prefix_check()
 requirements_osx_brew_lib_installed()
 {
   \typeset brew_lib_prefix
-  brew list -1 | __rvm_grep "^${1}$" >/dev/null &&
+  # Test for older versions of Homebrew before adding the `--formula` flag.
+  brew_version=$( brew --version | tee | head -n 1 | sed 's/[^0-9.]*\([0-9.]*\).*/\1/' )
```
miketheman, line 43: `tee` was necessary to prevent a broken pipe, since the brew command will return one line before proceeding to check other versions of casks. If there's a preferred way of doing this, happy to adapt.

pkuczynski (dismissed):
LGTM!

scripts/functions/requirements/osx_brew
DNA: I know the PR is merged, but if you can use AWK it gets cleaner:

```bash
 brew --version | awk 'NR==1 { print $2 }'
 ```

pkuczynski: Do you wanna fire up PR with this change?

DNA: Sure thing! I wasn't sure if it was a change worth enough to open a PR, but I'll make one now :)

pkuczynski: Does not hurt for sure :) Thanks!

PR #5005, opened 2020-11-17
Add TruffleRuby 20.3.0

eregon:
cc @havenwood, @pkuczynski


pkuczynski (approved):

eregon:
CI is passing, except the first job which was already failing and unrelated:
https://travis-ci.org/github/rvm/rvm/builds/742774481
This means I can't click `Merge pull request`.

I think let's add that job to allow failures since it fails anyway.

eregon:
@pkuczynski Could you re-approve since pushing a new commit removed the approval?

pkuczynski:
Sure!

pkuczynski (approved):

.travis.yml
```diff
@@ -59,6 +59,9 @@ matrix:
         - rake --version
         - ruby -S bundle --version
         - ruby -ropen-uri -e 'puts open("https://rubygems.org/") { |f| f.read(1024) }'
+  allow_failures:
```
pkuczynski, line 62: Why do we want this?

pkuczynski (changes requested):

.travis.yml
eregon: Otherwise I cannot merge the PR, as I explained in https://github.com/rvm/rvm/pull/5005#issuecomment-729604991.

eregon: In case it wasn't clear, that job has been failing [for ages](https://travis-ci.org/github/rvm/rvm/builds), and nobody seems to care to fix it, so there is no point to pretend it passes, and I think we can acknowledge it fails and have the CI result actually reflects if the PR breaks something.
Plus, it prevents PRs for contributors like me with write access.

pkuczynski: Make sense!

pkuczynski (approved):

PR #5009, opened 2020-12-08
Adds support for Ruby 3.0.0-preview2, changes Ruby 3 to .tar.gz

pcarn:
[Ruby 3.0.0-preview2 has been released.](https://www.ruby-lang.org/en/news/2020/12/08/ruby-3-0-0-preview2-released/)

For some reason there is not a `.tar.bz2` version of it: https://cache.ruby-lang.org/pub/ruby/3.0/

For now, I've changed Ruby 3 to install from `.tar.gz`, I'll keep an eye on the next version.

PR #5010, opened 2020-12-08
Support ruby 3.0.0 preview2

sqbell:
* Adds support for ruby 3.0.0 preview 2
* Switches to smaller `xz` files that are available since ruby 2.1.3


config/known
```diff
@@ -13,6 +13,7 @@
 [ruby-]2.6[.6]
 [ruby-]2.7[.2]
 [ruby-]3[.0.0-preview1]
```
havenwood, line 15: It looks like this line should be removed since the next line replaces its function.

sqbell:
I think we can close this one as https://github.com/rvm/rvm/pull/5009/files fixes.

PR #5022, opened 2020-12-27
Use AWK to check Brew version

DNA:
The version check used on #4995 can be improved with a simpler AWK command, merging 3 commands into one.


pkuczynski (approved):

pkuczynski:
Thanks! :)

PR #5037, opened 2021-01-19
Add TruffleRuby 21.0.0

eregon:
https://github.com/oracle/truffleruby/releases/tag/vm-21.0.0


eregon:
CI is green, ready to merge, except for that one unrelated failure, which was already on master:
https://travis-ci.org/github/rvm/rvm/jobs/754678007

config/known
```diff
@@ -35,7 +35,7 @@ rbx-5[.0]
 rbx-head
 
 # TruffleRuby
-truffleruby[-20.3.0]
+truffleruby[-21.0.0]
```
pkuczynski, line 38: Should we maybe do:

```
truffleruby-20[.3.0]
truffleruby[-21.0.0]
```

In case someone would like to stick to 20.x?

eregon: They can still `rvm install truffleruby-20.3.0`, so I think there is no need.
TruffleRuby 20.x is no longer maintained, only the latest release is.

pkuczynski: Well, that's what we do for other rubies, always providing maximum minor for every major version. Even when they not maintained. Have a look here: https://github.com/rvm/rvm/blob/master/config/known#L2

eregon: We didn't do it for all the [previous releases](https://github.com/postmodern/ruby-versions/blob/master/truffleruby/versions.txt), and I don't think it makes sense to list them here for a shortcut.
Old versions can still be installed, but only the latest version is the [stable version](https://github.com/postmodern/ruby-versions/blob/master/truffleruby/stable.txt).

There is no such thing as compatibility between 20.1 and 20.2, it's the same as between 20.3 and 21.0, i.e., the first number is just the year, and I think it doesn't make sense to pin to that (or have a shortcut for it). E.g., we might update to Ruby 3 in 21.2.
In other words, if users want to pin to a version, they should specify it in full, like `truffleruby-20.3.0`, not `truffleruby-20`.

pkuczynski: Got it, fair enough. Thanks for the explanaition

pkuczynski (approved):

PR #5040, opened 2021-01-21
GitHub Actions

### LINKED ISSUE #5038: Try travis.com
It seems the CI currently uses travis-ci.org, e.g., https://travis-ci.org/github/rbenv/ruby-build/builds/755249540

As they say:
> Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.

Maybe migrating to travis-ci.com would make it a bit faster? Right now it's horribly long to wait for.
Probably migrating to another CI is a better long-term solution.

pedrofurtado:
Closes #5038.


pedrofurtado:
@pkuczynski @mpapis  When you have time, can you help me writing this file? I started some draft, migrating parts of the `.travis.yml` to `.github/workflows/all.yml`.

Later, we can define a pattern for names, build steps and so on 🤝

.github/workflows/gpg_signature.yml
```diff
@@ -0,0 +1,19 @@
+name: GPG signature
+on: [push]
+jobs:
+  build:
+    name: GPG signature
+    strategy:
+      matrix:
```
pkuczynski, line 7: I think there is no value running this in matrix. One OS should be enough and I would use some smaller image than ubuntu...

eregon: ubuntu is what's available on GH Actions, and the matrix is just one job, but yeah it could be simplified to not use a matrix for clarity.

eregon (dismissed):
This looks good to me.
I would suggest to merge it (see https://github.com/rvm/rvm/issues/5038#issuecomment-883576451) and then expand it to e.g. run on macOS too.

We'll want to squash all commits though.

.github/workflows/gpg_signature.yml
pedrofurtado: The matrix allow us to add ruby versions to be included in tests. For a while, only ruby 2.7 is defined, but we can add later 3.0, 3.1, 2.6 and so on 🍻

pedrofurtado:
Github actions running on Ubuntu and MacOS 🎉 

The commits can be squashed by administrators, when merging the PR (we just need to select the option "squash and merge")

pkuczynski:
Looks good to me too! And its much faster. I will give it a go...

pkuczynski:
Thank you @pedrofurtado for putting your time into this!

PR #5083, opened 2021-05-26
Fix hashes for TruffleRuby 21.1.0

eregon:
* Unfortunately the original builds were based on incorrect JDK versions,
  these new hashes represent the correct builds based on JDK11.


eregon:
@havenwood Hello! Could you review and merge this?
Otherwise I guess rvm users will get warnings/errors of mismatched checksums until then.

pkuczynski (approved):

eregon:
Thanks for merging!

PR #5104, opened 2021-07-20
Improve GitHub Actions

eregon:
Follow-up of #5040
cc @pedrofurtado

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -14,8 +14,10 @@ jobs:
       - uses: actions/checkout@v2
       - run: gem install tf -v '>=0.4.1'
       - run: ./install
-      - name: rvm-test/fast/* rvm-test-rvm1/*
-        run: source ~/.rvm/scripts/rvm && tf --text rvm-test/fast/* rvm-test-rvm1/*
+      - name: rvm-test/fast/*
+        run: source ~/.rvm/scripts/rvm && tf --text rvm-test/fast/*
+      - name: rvm-test-rvm1/*
+        run: source ~/.rvm/scripts/rvm && tf --text rvm-test-rvm1/*
       - name: named_ruby_and_gemsets_comment_test
         run: source ~/.rvm/scripts/rvm && tf --text rvm-test/long/named_ruby_and_gemsets_comment_test.sh
       - name: truffleruby_comment_test
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -3,6 +3,7 @@ on: [push, pull_request]
 jobs:
   tests:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
         ruby: [2.7]
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -6,12 +6,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        ruby: [2.7]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: 2.7
       - uses: actions/checkout@v2
       - run: gem install tf -v '>=0.4.1'
       - run: ./install
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -13,15 +13,11 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - uses: actions/checkout@v2
-      - name: Specs
-        run: |
-          gem install tf -v '>=0.4.1'
-          hash -r
-          ./install
-          git submodule init
-          git submodule update
-          source ~/.rvm/scripts/rvm
-          rvm reload
-          tf --text "rvm-test/fast/* rvm-test-rvm1/*"
-          tf --text "rvm-test/long/named_ruby_and_gemsets_comment_test.sh"
-          tf --text "rvm-test/long/truffleruby_comment_test.sh"
+      - run: gem install tf -v '>=0.4.1'
+      - run: ./install
+      - name: rvm-test/fast/* rvm-test-rvm1/*
+        run: source ~/.rvm/scripts/rvm && tf --text "rvm-test/fast/* rvm-test-rvm1/*"
+      - name: named_ruby_and_gemsets_comment_test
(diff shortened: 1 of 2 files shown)


.github/workflows/specs.yml
```diff
@@ -3,24 +3,20 @@ on: [push, pull_request]
...
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: 2.7
       - uses: actions/checkout@v2
-      - name: Specs
-        run: |
-          gem install tf -v '>=0.4.1'
-          hash -r
-          ./install
-          git submodule init
```
pkuczynski, line 20: We do not need this anymore?

.github/workflows/specs.yml
```diff
@@ -3,24 +3,22 @@ on: [push, pull_request]
...
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        ruby: [2.7]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
```
pkuczynski, line 13: I think doing this in matrix makes sense, as we could test different versions of ruby...

.github/workflows/specs.yml
eregon: 81b56ae4747aa7f2d3056fa92a7df1ad18095338
>  git submodule commands are no longer needed, there are no submodules in the rvm repo anymore

.github/workflows/specs.yml
```diff
@@ -3,24 +3,22 @@ on: [push, pull_request]
...
       - uses: actions/checkout@v2
-      - name: Specs
-        run: |
-          gem install tf -v '>=0.4.1'
-          hash -r
-          ./install
-          git submodule init
-          git submodule update
-          source ~/.rvm/scripts/rvm
-          rvm reload
```
pkuczynski, line 23: Are you sure `rvm reload` is not needed?

.github/workflows/specs.yml
eregon: Do the tests even use the current version of Ruby?
One important thing to note here is the `ruby` in PATH is not a Ruby built or installed by RVM.
I would think for tests you would want to test a RVM Ruby.
I can remove this change anyway.

.github/workflows/specs.yml
eregon: GH Actions steps are all isolated, they don't run in the same bash process.

eregon:
It seems tests fail now.
I just removed quotes around the test names, which I believe is necessary to execute all tests.
https://github.com/rvm/rvm/runs/3117715636
Not sure how come the tests passed before in GH Actions, because having the quotes triggers an error: https://github.com/rvm/rvm/runs/3117695652

It seems the original run did not run much things, in fact it seems we almost only see the output of `truffleruby_comment_test.sh` and almost no output from the rest:
https://github.com/rvm/rvm/runs/3117402756?check_suite_focus=true

eregon:
https://github.com/rvm/rvm/runs/3117402756?check_suite_focus=true#step:4:45 so the error due the quotes was there before, but it seems that didn't make the whole step fail, probably because of other shell commands after that succeeded.

.github/workflows/specs.yml
pkuczynski: I remember we talked about it, but dont remember if we executed this change or not?

eregon:
This seems the first error:
https://github.com/rvm/rvm/pull/5104/checks?check_run_id=3117756898#step:6:232
```
 Required ruby-2.3.4 is not installed - installing.
Error running 'requirements_debian_libs_install libssl1.0-dev',
please read /home/runner/.rvm/log/1626811406_ruby-2.3.4/package_install_libssl1.0-dev.log
Requirements installation failed with status: 100.
```

libssl1.0-dev no longer exist on recent Ubuntu (OpenSSL 1.0 is EOL), and I think even less so on GitHub Actions runners.
So I think the tests needs to be updated to use a less ancient Ruby.
Ruby 2.4 can already handle OpenSSL 1.1.
Ruby 2.6 would be the current security maintenance Ruby.

pkuczynski:
> It seems tests fail now.

Ooops. I can see there are a lot of issues with installing ruby. Would you find some time to look into this?

.github/workflows/specs.yml
pkuczynski: Yeah, you're probably right!

eregon:
> Ooops. I can see there are a lot of issues with installing ruby. Would you find some time to look into this?

It's getting late here so actually I would be happy if somebody else takes over.
We could either for instance merge this and then try to make the tests pass.

Is there a way to make [tf](https://github.com/rvm/tf) stop at the first failure?
It's quite unclear from the output what actually fails, I just guessed by searching for `error`.
Maybe `TF_DEBUG=1` is helpful, that would be worth a try.

.github/workflows/specs.yml
eregon: Yes it's done, there is no `.gitmodules` at the root of the repo anymore

pkuczynski:
> It's getting late here so actually I would be happy if somebody else takes over.

I dont think there will be anybody else available anytime soon, I am afraid

> We could for instance merge this and then try to make the tests pass.

Yes, let's do that. Its better to have true failing tests than false-positive ones I guess...?

> Is there a way to make [tf](https://github.com/rvm/tf) stop at the first failure?
> It's quite unclear from the output what actually fails, I just guessed by searching for `error`.

I don't know. @mpapis was the author. He is travelling now, so I doubt he will be able to look into this anytime soon...

eregon:
FWIW on macOS there are less failures:
https://github.com/rvm/rvm/pull/5104/checks?check_run_id=3117756946
> ##### Processed commands 621 of 621, success tests 402 of 434, failure tests 32 of 434.
vs Ubuntu
https://github.com/rvm/rvm/pull/5104/checks?check_run_id=3117756898
>  ##### Processed commands 621 of 621, success tests 342 of 434, failure tests 92 of 434.

Maybe due to openssl 1.0 working there (if Homebrew still has it)

eregon:
> Yes, let's do that. Its better to have true failing tests than false-positive ones I guess...?

OK let's merge and I'll try to find some time to investigate further.
At least these changes should be a safe refactoring and the only semantic change is they actually run all tests and CI fails if tests fail.

pkuczynski (approved):

pkuczynski:
> OK let's merge and I'll try to find some time to investigate further.

That would be really awesome, thanks!

eregon:
Regarding the quotes, the issue is `tf` is pretty rigid about being given filenames ending in `_comment_test.sh`, and so it has to be not quoted and expanded before tf gets it.

PR #5106, opened 2021-07-20
Move all tests into a common location

pkuczynski:
Let's align all tests into one common location...

I am not sure what's the reason behind splitting `rvm1` and `fast`. I asked @mpapis and will follow up with a necessary PR once he will get back to me.


mpapis:
Fast and long tests were supposed to be implementation independent tests for both rvm1 and rvm2, as the V2 effort is not progressing you could merge them back together with rvm1 - as it's tests with details specific to rvm1.

eregon (dismissed):
Having all tests under `tests/` (or `test/`) seems nice :+1:

eregon:
It seems the set of required jobs does not match the actual actions, `CI (macos-latest, 2.7)` vs `CI / tests (ubuntu-latest)` it seems, probably because I removed `2.7` from the matrix.
As a result I can't merge this PR, and I'll wait for this PR to try to fix the CI to avoid conflicts.

pkuczynski:
> Having all tests under `tests/` (or `test/`) seems nice 👍

I used `tests` in the plural, as we use the plural form for other folders

pkuczynski:
> Fast and long tests were supposed to be implementation independent tests for both rvm1 and rvm2, as the V2 effort is not progressing you could merge them back together with rvm1 - as it's tests with details specific to rvm1.

I merged `rvm1` into `fast`. Thanks for promptly reply @mpapis 

Is there any reason why do we run only 2 tests from `long` and not all @mpapis  @eregon?

pkuczynski:
@eregon I forced merged this one despite failing tests, so you could pick it up now and try to figure out what's failing. Thank you for your efforts!

eregon:
I noticed now there are actually duplicate test files in master, e.g.:
https://github.com/rvm/rvm/blob/master/tests/long/truffleruby_comment_test.sh
https://github.com/rvm/rvm/blob/master/rvm-test/long/truffleruby_comment_test.sh

Should we `git rm -r rvm-test`?

eregon:
I'll make a PR to remove the duplicate `rvm-test` dir after https://github.com/rvm/rvm/pull/5177 is merged (to avoid conflicts)

PR #5109, opened 2021-07-21
Fix GitHub Actions CI

eregon:
Follow-up of #5104, and ref #5038.

CI passes on Linux now: https://github.com/eregon/rvm/actions/runs/1053563216
Let's see how it goes on macOS.

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -9,6 +9,8 @@ jobs:
           - ubuntu-latest
           - macos-latest
     runs-on: ${{ matrix.os }}
+    env:
+      TERM: ansi
     steps:
       - uses: ruby/setup-ruby@v1
         with:
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -15,6 +15,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: 2.7
+          bundler: none
       - uses: actions/checkout@v2
       - run: gem install tf -v '>=0.4.1'
       - run: ./install
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -12,10 +12,6 @@ jobs:
     env:
       TERM: ansi
     steps:
-      - uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: 2.7
-          bundler: none
       - uses: actions/checkout@v2
       - run: ./install
       - run: source ~/.rvm/scripts/rvm && rvm use 2.4.0 --install --default && gem install tf -v '>=0.4.1'
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -17,8 +17,8 @@ jobs:
           ruby-version: 2.7
           bundler: none
       - uses: actions/checkout@v2
-      - run: gem install tf -v '>=0.4.1'
       - run: ./install
+      - run: source ~/.rvm/scripts/rvm && rvm use 2.4.0 --install --default && gem install tf -v '>=0.4.1'
       - name: tests/fast/*
         run: source ~/.rvm/scripts/rvm && tf --text tests/fast/*
       - name: named_ruby_and_gemsets_comment_test
--- a/tests/fast/alias_comment_test.sh
+++ b/tests/fast/alias_comment_test.sh
@@ -1,6 +1,6 @@
 source "$rvm_path/scripts/rvm"
 
-rvm use 2.3.4 --install
+rvm use 2.6.6 --install
 rvm use 2.4.0 --install
 rvm gemset create test_gemset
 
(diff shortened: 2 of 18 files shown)


eregon:
@pkuczynski Could you adjust the `Required` jobs in the protected branch settings to match the new job names (no `, 2.7` in it)?

eregon:
There are no binaries for macOS 10.15 at https://rvm_io.global.ssl.fastly.net/binaries/osx/
So on macOS all Rubies are built from source, that's going to be pretty slow (e.g., https://github.com/rvm/rvm/runs/3124143229?check_suite_focus=true took 20min and only execute a subset of the tests).

eregon:
Does RVM need precompiled Rubies to be built by RVM?
If not, maybe we could configure RVM in the macOS CI to use binaries from https://github.com/ruby/ruby-builder/releases/latest.

eregon:
There are still some errors on macOS:
##### Processed commands 991 of 991, success tests 618 of 632, failure tests 14 of 632.
https://github.com/rvm/rvm/pull/5109/checks?check_run_id=3127115584#step:5:1466
https://github.com/rvm/rvm/pull/5109/checks?check_run_id=3127115584#step:5:1682
https://github.com/rvm/rvm/pull/5109/checks?check_run_id=3127115584#step:5:1778

I won't have time to investigate those, and also I don't have a macOS machine to debug them.

Still the Linux CI is green now, so I'd suggest to merge this, and not make the macOS CI job `Required` until it's fixed.

BrianHawley:
@eregon it looks like the `rvm mount -r ruby-2.6.6.tar.bz2` test failed in macOS, then all the 2.6.6 tests after that failed until 2.6.6 got fixed in a later test. I have a macOS machine to debug, and might have time to do so later today or later this week.

pkuczynski:
> Does RVM need precompiled Rubies to be built by RVM?
> If not, maybe we could configure RVM in the macOS CI to use binaries from https://github.com/ruby/ruby-builder/releases/latest.

No, this is not required. We could technically take a binary ruby from any location. Correct @mpapis?

pkuczynski:
> @pkuczynski Could you adjust the `Required` jobs in the protected branch settings to match the new job names (no `, 2.7` in it)?

Can do that once this PR is merged...

pkuczynski:
> There are no binaries for macOS 10.15 at https://rvm_io.global.ssl.fastly.net/binaries/osx/
> So on macOS all Rubies are built from source, that's going to be pretty slow (e.g., https://github.com/rvm/rvm/runs/3124143229?check_suite_focus=true took 20min and only execute a subset of the tests, https://github.com/eregon/rvm/runs/3127108618?check_suite_focus=true took 7min30 to build Ruby 2.4.0).

It's a known problem :/ There was an issue with building movable macOS binaries which we never really solved :(

pkuczynski:
Merging this in, as it fixes `ubuntu` tests. Hoping @BrianHawley will find some time to fix macOS

mpapis:
> > Does RVM need precompiled Rubies to be built by RVM?
> > If not, maybe we could configure RVM in the macOS CI to use binaries from https://github.com/ruby/ruby-builder/releases/latest.
> 
> No, this is not required. We could technically take a binary ruby from any location. Correct @mpapis?
> 
> 

It might be fine, in the past Ruby had issues on osx when compiled without some flags, some of the env vars were not respected, and incompatibility conflicts arraised because of using wrong paths, we would need to check an rails app with few Rubie's installed in rvm1, on system and in default path from Ruby-build, i was reporting those issues to ruby, some of the patches were incorporated some weren't

eregon:
To be clear I was suggesting to use [those binaries](https://github.com/ruby/ruby-builder/releases/latest) only for speeding up CI, not for RVM users.
Those binaries cannot be moved, movable Rubies have [many issues](https://github.com/ruby/setup-ruby/issues/98#issuecomment-721233467) (especially older versions), as I guess you are already aware.

pkuczynski:
> To be clear I was suggesting to use [those binaries](https://github.com/ruby/ruby-builder/releases/latest) only for speeding up CI, not for RVM users.
> Those binaries cannot be moved, movable Rubies have [many issues](https://github.com/ruby/setup-ruby/issues/98#issuecomment-721233467) (especially older versions), as I guess you are already aware.

That makes sense to me. It's horribly slow at the moment :( But how would you like to do that without changing RVM?

eregon:
Is it possible to tell RVM to use a binary at a given location?
For instance 
```yml
- uses: ruby/setup-ruby@v1
  with:
    ruby-version: 2.6.6
```
will ensure there is `/Users/runner/hostedtoolcache/Ruby/2.6.6/x64` (and `bin/ruby` under that) available (which can be queried via the `ruby-prefix` output of that action).
Can RVM then "add/adopt" this Ruby without needing to move it?

PR #5101, opened 2021-07-16
Fix Termux

BrianHawley:
Changes proposed in this pull request:
* Remove unnecessary and unsupported dependencies.
* No getent, so use echo $SHELL instead.
* Add Windows:Mingw and Linux:Termux support to __rvm_get_user_shell
  duplicate declaration in warning_zsh too.

Fixes #5003, #5048, #5057

PR #5110, opened 2021-07-22
Fix Termux requirements

BrianHawley:

RVM on Termux was waiting during its requirements install for
user input in a background prompt. The pkg command, based on apt,
had no option to make it accept that prompt automatically, so I
switched the operations to use apt-get, which had such options.

Also, went through the requirements lists and made the lists more
specific for various ruby types. Some ruby types won't work.


BrianHawley:
@pkuczynski this is the follow-up PR for #5101.

BrianHawley:
@pkuczynski I resolved the conflict.

pkuczynski (approved):
Great work! Thanks!

Issue #5114, opened 2021-07-22
rvm install ruby-2 tries to install truffleruby


BrianHawley:
## Description

If you try to install ruby-2, it installs truffleruby-21.2.0 (the latest truffleruby with the string "ruby-2" in the middle of it, as of the time of writing).

## Steps to reproduce

 1. `rvm get master`
 2. `rvm install ruby-2`

## Expected behavior

Expected ruby-2.7.4 to be installed.

## Actual behavior

```
Checking requirements for osx.
Installing requirements for osx.
Updating system.........................
Installing required packages: openssl.
Certificates bundle '/usr/local/etc/openssl@1.1/cert.pem' is already up to date.
Requirements installation successful.
truffleruby-21.2.0 - #downloading truffleruby-21.2.0-macos-amd64, this may take a while depending on your connection...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   643  100   643    0     0   2115      0 --:--:-- --:--:-- --:--:--  2108
100  113M  100  113M    0     0  23.2M      0  0:00:04  0:00:04 --:--:-- 25.8M
truffleruby-21.2.0 - #extracting truffleruby-21.2.0-macos-amd64 to /Users/bhawley/.rvm/src/truffleruby-21.2.0.....
truffleruby-21.2.0 - #installing to /Users/bhawley/.rvm/rubies/truffleruby-21.2.0.
truffleruby-21.2.0 - #compiling c-extensions............
truffleruby-21.2.0 - #making binaries executable...
truffleruby-21.2.0 - #gemset created /Users/bhawley/.rvm/gems/truffleruby-21.2.0@global
truffleruby-21.2.0 - #importing gemset /Users/bhawley/.rvm/gemsets/truffleruby/global.gems..................................
truffleruby-21.2.0 - #generating global wrappers........
truffleruby-21.2.0 - #gemset created /Users/bhawley/.rvm/gems/truffleruby-21.2.0
truffleruby-21.2.0 - #importing gemsetfile /Users/bhawley/.rvm/gemsets/default.gems evaluated to empty gem list
truffleruby-21.2.0 - #generating default wrappers........
RVM gem rubygems-bundler is not installed by default for TruffleRuby.
Use 'bundle exec' instead when needed. See rvm/rvm#4765.
```

[...]

PR #5115, opened 2021-07-22
Install ruby-head with ruby 3

### LINKED ISSUE #5112: ruby-head won't install with ruby 3
## Description

I want to be able to install ruby-head if I only have ruby 3 installed, but it looks for ruby 2, and tries to install truffleruby if it doesn't find it (which is not available on all platforms).

## Steps to reproduce

 1. Install rvm and get master, if it isn't installed.
 2. If any MRI ruby 2 or 1.9 is installed, uninstall it so it doesn't get chosen.
 3. `rvm install ruby-head`

## Expected behavior

If you don't have any MRI ruby installed, it should install the latest MRI ruby release (3.0.2 at the time of writing), then attempt to install ruby-head. If you have a MRI ruby 3 installed, it should use it to attempt to install ruby-head.

## Actual behavior

It doesn't use ruby 3 to install ruby-head. Instead, it tries to install the latest truffleruby (which fails on the platform I tested) rather than the latest ruby.

I don't have command logs because I did this on a platform where it would be awkward to retrieve them. However, I tracked down the cause and will be making a PR.

BrianHawley:
* Adds ruby 3 as an allowed ruby to install ruby-head with.
* Be more specific about checking MRI rubies to exclude non-MRI rubies with the MRI ruby prefix somewhere in the version string.

Fixes #5112.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Bug fixes
 
 * Remove unsupported libclang, libclang-dev, openssl-dev and zlib-dev from Termux requirements, and fix installation by not calling getent [\#5101](https://github.com/rvm/rvm/pull/5101)
+* Allow ruby-head to be installed using ruby 3, not truffleruby [\#5115](https://github.com/rvm/rvm/pull/5115)
 
 #### New interpreters
 
--- a/scripts/functions/build
+++ b/scripts/functions/build
@@ -306,10 +306,10 @@ __rvm_mono_env()
 __rvm_mri_rubies()
 (
   # find on bsd does not have -not, we need to use \!
-  \typeset versions="${1:-"ruby-1.9|ruby-2"}"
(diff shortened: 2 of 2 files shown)


BrianHawley:
Note that this fixes the problem of truffleruby being used to create ruby-head if MRI ruby 3 is installed, but no MRI ruby 2 or 1.9 installed.

It also temporarily solves the problem of truffleruby being used or installed before installing ruby-head if no MRI ruby is installed already, but this is only because it searches for the latest ruby-3 in that case. Its previous behavior was to install truffleruby 21.2.0 in that case because the string "ruby-2" is inside some of the truffleruby version strings, because of #5114. There is currently no truffleruby version with "ruby-3" inside the version string, so ruby-3.0.2 will currently be installed. However, if truffleruby 30 comes out before #5114 is fixed, we will have the problem on fallback again. In any case, that problem is not in related code.

pkuczynski (approved):

pkuczynski:
Thanks!

Issue #4937, opened 2020-05-18
Unable to build movable macOS binary on Catalina 10.15.4


hurricup:
Trying to build a movable ruby:
```
% rvm install 2.7.1 --movable --autolibs=homebrew
It is not yet supported to build movable rubies with 'homebrew',
please install SMF and switch autolibs to it (make sure to follow any displayed instructions):

    curl -L https://get.smf.sh | sh
    rvm autolibs smf
```

Smf link is dead and i could not find smf manually quickly. Probably it's dead too?

What is the next step. And why other options not working, like `fail` or `homebrew`?

rvm info: 
```
ruby-2.6.5:
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance

  system:
    uname:        "Darwin admins-Mac-mini.local 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64"
    name:         "OSX"
    version:      "10.15"
    architecture: "x86_64"
    bash:         "/bin/bash => GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)"
    zsh:          "/bin/zsh => zsh 5.7.1 (x86_64-apple-darwin19.0)"
    remote_path:  "osx/10.15/x86_64"
    xcode:        ""

  rvm:
    version:      "1.29.10 (latest)"
    updated:      "4 minutes 4 seconds ago"
    path:         "/Users/user/.rvm"
    autolibs:     "[homebrew] Allow RVM to use package manager if found, install missing dependencies, install package manager (only OS X)."

  ruby:
    interpreter:  "ruby"
    version:      "2.6.5p114"
    date:         "2019-10-01"
    platform:     "x86_64-darwin19"
    patchlevel:   "2019-10-01 revision 67812"
    full_version: "ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin19]"

  homes:
    gem:          "/Users/user/.rvm/gems/ruby-2.6.5"
    ruby:         "/Users/user/.rvm/rubies/ruby-2.6.5"

  binaries:
    ruby:         "/Users/user/.rvm/rubies/ruby-2.6.5/bin/ruby"
    irb:          "/Users/user/.rvm/rubies/ruby-2.6.5/bin/irb"
[...]

PR #5119, opened 2021-07-30
Fix CI for macOS

### LINKED ISSUE #5118: Prepare and mount test not movable, and fails on macOS
## Description

tests/fast/remote_comment_test.sh uses a ruby that is not movable, and doesn't work at all on macos because of #4937.

Installing Ruby so it can be used with `rvm prepare` is supposed to use the `--movable` option, though this option doesn't do enough on Ubuntu that you can't run `rvm prepare` and `rvm mount` without it being specified. So, on Ubuntu, the existing test passes by accident.

CI tests for macOS run with Homebrew as the package manager. The `--movable` option on macOS only works with the SMF package manager. However, there is no longer any documented way to install the SMF package manager at all, as its installation and documentation servers are not running anymore, and the project looks abandoned.

The test also uses a Ruby version that is used in later tests, so after the test fails on macOS, later tests that use the same version (and the default gemset) also fail because the installation of that version is corrupted.

Also, installing a ruby with `--movable` is supposed to compile, even when not on macOS, so the test arguably should be in the tests/long directory.

## Steps to reproduce

Running on macOS, with any package manager other than SMF.

 1. `rvm install ruby-2.6.8 --movable`

## Expected behavior

It should install.

## Actual behavior

```
Required ruby-2.6.8 is not installed - installing.
It is not yet supported to build movable rubies with '4',
please install SMF and switch autolibs to it (make sure to follow any displayed instructions):

    curl -L https://get.smf.sh | sh
    rvm autolibs smf

Gemset '' does not exist, 'rvm ruby-2.6.8 do rvm gemset create ' first, or append '--create'.
```

That last bit is probably why the default gemset tests later on fail.

## Environment info

Normal CI for macOS.

BrianHawley:
* Move remote_comment_test.sh out of fast and don't run it on macOS.
* Make sure that the Ruby used in the prepare test is --movable.
* Update the Darwin rvm_movable_flag version check to include Ruby 3.

Fixes #5118.

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -24,4 +24,5 @@ jobs:
         if: ${{ !contains(matrix.os, 'macos') }}
         run: source ~/.rvm/scripts/rvm && tf --text tests/long/ruby_prepare_mount_comment_test.sh
       - name: truffleruby_comment_test
+        if: ${{ !contains(matrix.os, 'macos') }}
         run: source ~/.rvm/scripts/rvm && tf --text tests/long/truffleruby_comment_test.sh
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -18,11 +18,14 @@ jobs:
       - name: tests/fast/*
         run: source ~/.rvm/scripts/rvm && tf --text tests/fast/*
       - name: named_ruby_and_gemsets_comment_test
+        # works on local, but fails in CI, needs to be investigated
         if: ${{ !contains(matrix.os, 'macos') }}
         run: source ~/.rvm/scripts/rvm && tf --text tests/long/named_ruby_and_gemsets_comment_test.sh
       - name: ruby_prepare_mount_comment_test
+        # https://github.com/rvm/rvm/issues/4937
         if: ${{ !contains(matrix.os, 'macos') }}
         run: source ~/.rvm/scripts/rvm && tf --text tests/long/ruby_prepare_mount_comment_test.sh
       - name: truffleruby_comment_test
+        # works on local, but fails in CI, needs to be investigated
         if: ${{ !contains(matrix.os, 'macos') }}
         run: source ~/.rvm/scripts/rvm && tf --text tests/long/truffleruby_comment_test.sh
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -18,9 +18,10 @@ jobs:
       - name: tests/fast/*
         run: source ~/.rvm/scripts/rvm && tf --text tests/fast/*
       - name: named_ruby_and_gemsets_comment_test
+        if: ${{ !contains(matrix.os, 'macos') }}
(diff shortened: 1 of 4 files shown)


BrianHawley:
@pkuczynski this temporarily works around #4937 by not trying to prepare and mount on macOS.

BrianHawley:
tests/long/named_ruby_and_gemsets_comment_test.sh and tests/long/truffleruby_comment_test.sh pass on my machine, but my machine isn't CI. I can't easily diagnose failures with no output (maybe later).

pkuczynski (dismissed):

pkuczynski:
Thank you so much @BrianHawley! Maybe we should add a comments to the excluded tests on macos explaining why they are excluded or pointing to github issue? What do you think?

pkuczynski (approved):

pkuczynski:
I posted some comments, what you think?

BrianHawley:
@pkuczynski the comments look good. I have no problem squash-merging it as is.

BrianHawley:
For future debugging reference:

The failures of the GA macOS run seem to be intermittent, and in one run, the initial setup code is what failed. Due to an apparent quirk in GA'a UI, it doesn't always show the full output in a failing run unless you're watching it happen, and I didn't watch it happen in these cases because I was busy doing other things (these are long runs). In any case, GA doesn't necessarily show the ongoing progress of rvm actions that are taking too long to run, dropping some output that shows up locally.

For instance, here is the complete output of the last failure:
```
Run source ~/.rvm/scripts/rvm && tf --text tests/long/truffleruby_comment_test.sh
  source ~/.rvm/scripts/rvm && tf --text tests/long/truffleruby_comment_test.sh
  shell: /bin/bash -e {0}
  env:
    TERM: ansi
Error: Process completed with exit code 1.
```
No idea what that one did.

pkuczynski:
Great observations! I spoke with @mpapis and he said that in the past only smf guaranteed that ruby build with `movable` flag would be actually movable between machines. So if we find an alternative which works, we have a green light to change to it...

.github/workflows/specs.yml
```diff
@@ -18,6 +18,14 @@ jobs:
...
       - name: named_ruby_and_gemsets_comment_test
+        # works on local, but fails in CI, needs to be investigated
+        if: ${{ !contains(matrix.os, 'macos') }}
         run: source ~/.rvm/scripts/rvm && tf --text tests/long/named_ruby_and_gemsets_comment_test.sh
+      - name: ruby_prepare_mount_comment_test
+        # https://github.com/rvm/rvm/issues/4937
+        if: ${{ !contains(matrix.os, 'macos') }}
+        run: source ~/.rvm/scripts/rvm && tf --text tests/long/ruby_prepare_mount_comment_test.sh
       - name: truffleruby_comment_test
+        # works on local, but fails in CI, needs to be investigated
```
eregon, line 29: Do you know what failed in CI?
I would expect this test to pass on macOS.
I'll try removing the `if` and see what happens.

PR #5148, opened 2021-11-07
Update binscripts/rvm-installer

tranch:
Add -q option to curl flags to avoid user custom definitions in .curlrc
file.

In some cases, user-defined configuration such as: -w "\n" will cause
the asc signature file to be changed and cannot be verified.


binscripts/rvm-installer
```diff
@@ -205,7 +205,7 @@ __rvm_curl()
   }
 
   typeset -a __flags
-  __flags=( --fail --location --max-redirs 10 )
+  __flags=( -q --fail --location --max-redirs 10 )
```
mpapis, line 208: I would rather go with:
```suggestion
  __flags=( $RVM_CURL_FLAGS --fail --location --max-redirs 10 )
```
this way you can overwrite any setting you like, but in most cases the `.curlrc` will be what users need like proxies, in case users have some nonstandard settings they can overwrite it.

mpapis (changes requested):

alivx (approved):

PR #5160, opened 2021-12-01
Add JRuby 9.2.20.1

headius:
Add JRuby 9.2.20.1


pkuczynski (approved):

pkuczynski (changes requested):
Missing CHANGELOG entry

headius:
The changelog requirement is pretty cumbersome, since independent changes will conflict with each other and I have to create the PR before I can update the changelog. It seems like this section could be auto-generated based on the added versions, couldn't it?

pkuczynski (approved):

pkuczynski:
Thanks!

> The changelog requirement is pretty cumbersome

You could also do all 3 versions in one PR, we would not mind.

PR #5149, opened 2021-11-07
Fixes #5096

mavaddat:
The keyserver `hkp://pool.sks-keyservers.net` has been deprecated. https://unix.stackexchange.com/questions/656205/sks-keyservers-gone-what-to-use-instead

Fixes # .

Changes proposed in this pull request:
*
*
*

Make sure that your pull request includes entry in the [CHANGELOG](CHANGELOG.md).

Issue #5096, opened 2021-07-02
gpgkeys: HTTP fetch error 6: Could not resolve host: pool.sks-keyservers.net


issuesreporting:
When following your official installation guide  (https://rvm.io/rvm/install)

 gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB

fails.

dgm:
This site is no longer available, see website:
http://sks-keyservers.net/
Update 2021-06-21: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

Ajmal:
gpg --keyserver keyserver.ubuntu.com --recv-key 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB

I'm using this one

jesperronn:
The workaround by Ajmal works for me (but only on my docker images, not on MacOS)

Would be great to have a workaround that works for  both unix and mac, so that the documentation could be updated.

See also https://github.com/SKS-Keyserver/sks-keyserver/issues/81 with info that old keyservers are no longer maintained

issuesreporting:
Yes, I know that URL is outdated. When documentation is updated this issue cases to occur as a consequence.

dgm:
The instruction on the website are still broken - How is anyone even able to install RVM right now?

x-yuri:
From what I can see [the recommended][a] and [the default keyserver][b] these days is `keys.openpgp.org`. This works for me:

```
$ gpg --keyserver keys.openpgp.org --recv-keys 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
```

Alternatively, you can use `keyserver.ubuntu.com`:

```
$ gpg --keyserver keys.openpgp.org --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
```

Apparently, `409B6B1796C275462A1703113804BB82D39DC0E3` can't be imported from `keys.openpgp.org`:

```
$ gpg --keyserver keys.openpgp.org --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
gpg: key 3804BB82D39DC0E3: no user ID
gpg: Total number processed: 1
```

Also, make sure you have `ca-certificates` installed, otherwise:

```
[...]

PR #5166, opened 2021-12-09
Key server to keys.openpgp.org

mavaddat:
Fixes # #5096
Fixes #5149

The keyserver `hkp://pool.sks-keyservers.net` has been deprecated. https://unix.stackexchange.com/questions/656205/sks-keyservers-gone-what-to-use-instead

### DIFF
--- a/binscripts/rvm-installer
+++ b/binscripts/rvm-installer
@@ -418,7 +418,7 @@ verify_package_pgp()
     log "\
 GPG signature verification failed for '$1' - '$3'! Try to install GPG v2 and then fetch the public key:
 
-    ${SUDO_USER:+sudo }${rvm_gpg_command##*/} --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+    ${SUDO_USER:+sudo }${rvm_gpg_command##*/} --keyserver hkp://keys.openpgp.org --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
 
 or if it fails:
 


phil-s:
See also https://github.com/rvm/rvm/pull/5150

The outdated (and not working) instructions are still on both https://rvm.io/rvm/install and https://rvm.io/rvm/security

binscripts/rvm-installer
```diff
@@ -418,7 +418,7 @@ verify_package_pgp()
     log "\
 GPG signature verification failed for '$1' - '$3'! Try to install GPG v2 and then fetch the public key:
 
-    ${SUDO_USER:+sudo }${rvm_gpg_command##*/} --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+    ${SUDO_USER:+sudo }${rvm_gpg_command##*/} --keyserver hkp://keys.openpgp.org --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
```
vogelbek, line 421: @mavaddat have you verified that keys.openpgp.org is working for you? I found I was only able to succeed with `hkp://pgp.mit.edu`. I can't recall if this is good practice or not, but would testing a :success response from the relevant server with the keys be a smart thing to add to the test suite? Additionally, might moving both server and keys to a config file reduce a bit of heartburn later?

eddietejeda:
The fix has been merged, but not deployed.

https://github.com/rvm/rvm-site/issues/348

mathieujobin:
link my last comment and PR here as well

https://github.com/rvm/rvm/issues/5291#issuecomment-1436506179
https://github.com/rvm/rvm-site/pull/365

binscripts/rvm-installer
mathieujobin: actually `keys.openpgp.org` is working for me as well, but not `pgp.mit.edu`.
but for consistency, I think we should stick with `keyserver.ubuntu.com` as I don't like having to change it at several places again... but I do prefer `keys.openpgp.org` actually...

PR #5198, opened 2022-03-31
Fix and update Rubinius modern versions, on modern systems

### LINKED ISSUE #5196: Proposal for fixing the Rubinius installation issues
Hello!

I've spent some time compiling Rubinius on my system yesterday, and I've identified several issues in RVM that prevent its compilation, for example:

- nonexisting clang package on newer Ubuntu versions
- unnecessarily old Ruby requirement for newer Rubinius versions
- incorrect build configure file for more recent Rubinius versions
- required patch to compile Rubinius with modern Ubuntu compilers
- etc.etc.

I was wondering if you're interested in me providing all the fixes. If so, what's the favourite issues/PR structure? E.g. one issue and one PR for each fix? Or keep a reference issue and multiple PRs? etc.etc.? I think a very granular approach would be best, especially because the issues are scattered in nature.

While Rubinius is abandoned (I guess), I think it's still a very valuable Ruby VM for research purposesl, so I think it'd be useful to have working RVM support!

64kramsystem:
Set of patches that fix and update Rubiniuns interpreters, on modern systems. The contents are detailed in each commit.

Among the changes, the default Rubinius version has been bumped to v5.0, currently the latest.

Verified on an Ubuntu 20.04 system, with Rubinius versions 4.7 and 5.0.

It'd be best if this could be tested on other systems. Performance hint: compile with `JOBS=$(nproc) rvm install rbx...`, in order to use all the cores during the main compilation stage (where nproc is not present, just use the number of hardware threads).

Closes #5196.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Skip non-functional prepare/mount tests on macOS [\#5119](https://github.com/rvm/rvm/pull/5119)
 * Fix detection for SUSE 15 [\#5132](https://github.com/rvm/rvm/pull/5132)
 * Avoid duplicate package instances when querying pkg-config [\#5172](https://github.com/rvm/rvm/pull/5172)
+* Fix building Ruby on Devuan 2 and newer by using libreadline-dev instead of libreadline6-dev [\#5214](https://github.com/rvm/rvm/pull/5214)
 
 #### New interpreters
 
--- a/scripts/functions/requirements/devuan
+++ b/scripts/functions/requirements/devuan
@@ -9,7 +9,7 @@ requirements_devuan_define_libgmp()
 
 requirements_devuan_define_libreadline()
 {
-  requirements_check libreadline6-dev
+  requirements_check libreadline-dev
 }
 
 requirements_devuan_define_libssl()


64kramsystem:
Closing for inactivity.

pkuczynski:
So is it working or not? If it does, we can merge it it...

64kramsystem:
Hello! I've tested it before, but on my system only (that's why I wrote it would have been best if it could be tested by someone on another sytem).

I'm going to rebase it just to be sure, and rerun my tests, then send a confirmation :+1:.

64kramsystem:
> So is it working or not? If it does, we can merge it it...

Hi! So,

I've made two changes:

1. converted the patchsets files to standard files; using a single file and symlinks made sense, but RVM doesn't fully support it;
2. I've changed the default Rubinius version to 4.15 (from 5.0), which is the most stable in my tests (and my system).

Regarding point 2, Rubinius is unfortunately not very stable on modern systems, but that's outside RVM's control.

I didn't rebase since there was little history since the PR was opened.

64kramsystem:
Closing for lack of communication - I've noticed that there was no follow up even on other trivial PR with fixes of mine that I've opened long ago.

Feel free to copy the code, I'm not fussy about ownership.

pkuczynski:
Hard to copy if you deleted everything. I had no time to look into your PR, as I have a little baby and not much time to work on open source, apologies. Seems that nobody else from the community was interested in rubinius, sorry about that... If you restore the code I can merge it as it is...

PR #5214, opened 2022-05-09
Use libreadline-dev instead of libreadline6-dev on Devuan 2

### LINKED ISSUE #5203: Use libreadline-dev instead of libreadline6-dev on Devuan
## Description

Building Ruby on Devuan fails because the expected requirement `libreadline6-dev` is missing. Only the first version of Devuan contained a `libreadline6-dev` actually, but since Devuan 2 there's only `libreadline-dev`.

## Steps to reproduce

 1. Install Devuan 2 or newer
 2. Install rvm stable 1.29.12 or rvm master
 3. rvm install 2.7.2 --disable-binary

## Expected behavior

Install Ruby using installed libreadline headers and libraries.

## Actual behavior

Install fails with message "Missing required packages: libreadline6-dev"

## Solution

I checked Devuan's version history on https://en.wikipedia.org/wiki/Devuan#Version_history and the package infos for "libreadline" on https://pkginfo.devuan.org, and the results are:

* Devuan 1 contains *both* `libreadline6-dev` and `libreadline-dev`, with `libreadline-dev` actually having a dependency on `libreadline6-dev`
* Devuan 2 and newer only contains `libreadline-dev`

This looks like the fix is simple and `libreadline-dev` can be used on all Devuan systems. In `scripts/functions/requirements/devuan`, change
```
requirements_check libreadline6-dev
```
to
```
requirements_check libreadline-dev
```
I made this change in my local RVM install on a Devuan 3 system and after that, building Ruby was successful.

jensberke:
Fixes #5203

### DIFF
--- a/scripts/functions/requirements/devuan
+++ b/scripts/functions/requirements/devuan
@@ -9,7 +9,7 @@ requirements_devuan_define_libgmp()
 
 requirements_devuan_define_libreadline()
 {
-  requirements_check libreadline6-dev
+  requirements_check libreadline-dev
 }
 
 requirements_devuan_define_libssl()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Skip non-functional prepare/mount tests on macOS [\#5119](https://github.com/rvm/rvm/pull/5119)
 * Fix detection for SUSE 15 [\#5132](https://github.com/rvm/rvm/pull/5132)
 * Avoid duplicate package instances when querying pkg-config [\#5172](https://github.com/rvm/rvm/pull/5172)
+* Fix building Ruby on Devuan 2 and newer by using libreadline-dev instead of libreadline6-dev [\#5214](https://github.com/rvm/rvm/pull/5214)
 
 #### New interpreters
 


scripts/functions/requirements/devuan
```diff
@@ -9,7 +9,7 @@ requirements_devuan_define_libgmp()
 
 requirements_devuan_define_libreadline()
 {
-  requirements_check libreadline6-dev
```
pkuczynski, line 12: Looks good in principle, but could you please make it dependent on the system version in a similar way like for example [here](https://github.com/rvm/rvm/blob/master/scripts/functions/requirements/debian#L74) or [here]()

pkuczynski (changes requested):
Thank you for the PR! Have a look at the requested change

scripts/functions/requirements/devuan
jensberke: I can that, but it's not really necessary, because all Devuan versions contain `libreadline-dev`, see https://pkginfo.devuan.org/cgi-bin/policy-query.html?c=package&q=libreadline-dev
`libreadline6-dev` only existed in Devuan's first version ("jessie"), being a dependency of `libreadline-dev`, i.e.  `libreadline-dev` and `libreadline6-dev` were the same there.

pkuczynski: Indeed! Thanks for explanation :)

pkuczynski (approved):

PR #5234, opened 2022-07-21
Use openssl on Void Linux

cherez:
We require LibreSSL on Void Linux, but Void dropped Libre to return to OpenSSL last year:
https://voidlinux.org/news/2021/02/OpenSSL.html

Switching the dependency to OpenSSL.


pkuczynski:
Could you please add entry in the CHANGELOG.md?

pkuczynski (approved):

pkuczynski (approved):

pkuczynski:
Thanks!

PR #5269, opened 2022-11-23
JRuby 9.4.0.0

headius:
Adds JRuby 9.4.0.0.


headius:
I can't tell what failed in those jobs but it does not seem related to JRuby.

pkuczynski:
Could you also update `config/known`. I noticed that we forgot to update it for normal ruby too, so separate PR for this would be also welcomed!

headius:
Looks like #5270 may be more complete? If so that one is just fine to merge.

havenwood:
Congrats on JRuby 9.4.0.0 release!

Closing since #5270 was merged, as suggested.

PR #5292, opened 2023-01-13
Don't warn about empty GEM_PATH

larskanis:
The `GEM_PATH` is set to an empty string by bundler when it is used per
```
  bundle install --path vendor/bundle
```
This is intentionally and checked in bundler's specs. See: https://github.com/rubygems/rubygems/blob/77686d0c4c8ffdbcbb8631878eebd0651ed8f90c/bundler/spec/bundler/bundler_spec.rb#L173-L177

However rvm then shows a warning, which is very misleading, since it points to an issue, which is not present at all. Also the warning can turn into a hard error, when the output is used from a subshell. For instance libusb's "make install" fails then, because of "garbage" in the shell output.

> Make sure that your pull request includes entry in the [CHANGELOG](CHANGELOG.md).

This often leads to merge conflicts, so that I didn't add it now. If this PR is OK, then I'll add a CHANGELOG entry.

### DIFF
--- a/scripts/rvm
+++ b/scripts/rvm
@@ -224,11 +224,10 @@ then
         fi
       elif
         [[ "${__path_to_ruby}" == "${rvm_path}"* ]] &&
-        [[ -z "${GEM_HOME:-}" || -z "${GEM_PATH:-}" ]]
+        [[ -z "${GEM_HOME:-}" ]]
       then
         echo "
-Warning: PATH set to RVM ruby but GEM_HOME and/or GEM_PATH not set, see:
-    https://github.com/rvm/rvm/issues/3212
+Warning: PATH set to RVM ruby but GEM_HOME not set.
 " >&2
         if
           [[ -n "${SUDO_USER:-}" ]]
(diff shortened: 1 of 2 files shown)


scripts/rvm
```diff
@@ -224,10 +224,10 @@ then
         fi
       elif
         [[ "${__path_to_ruby}" == "${rvm_path}"* ]] &&
-        [[ -z "${GEM_HOME:-}" || -z "${GEM_PATH:-}" ]]
+        [[ -z "${GEM_HOME:-}" ]]
       then
         echo "
-Warning: PATH set to RVM ruby but GEM_HOME and/or GEM_PATH not set, see:
```
pkuczynski, line 230: The fix about GEM_PATH could be valid, but the message would have to be different, as it now points to an issue 3212, which has nothing to do with GEM_PATH.

@mpapis what do you think?

larskanis: Another option could be to warm only on unset GEM_PATH, but to avoid the warning if it's set but empty.

To be honest, I don't fully understand the issue #3212 that is warned about, although I use rvm since many years. There is neither a good description of the impact of the problem nor how it can be solved.

pkuczynski: I haven't read it in detail. I think it would be enough to remove the reference to the issue and we can merge it in.

larskanis: OK, I removed the reference to the issue, added a CHANGELOG entry and fixed a URI typo in another entry.

pkuczynski (approved):

PR #5300, opened 2023-02-01
Add JRuby 9.3.10.0

headius:
Adds JRuby 9.3.10.0.


pkuczynski:
Can you also update `config/known` please?

headius:
> Can you also update `config/known` please?

Sure. Is that a new file? I don't think I updated that ever before.

pkuczynski:
> Sure. Is that a new file? I don't think I updated that ever before.

No, it was always there. I probably forgot to give you this feedback before, sorry...

headius:
No worries, it's done!

pkuczynski (approved):

PR #5336, opened 2023-04-01
New rubies 2.7.8, 3.0.6, 3.1.4, 3.2.2

hurricup:
New rubies 2.7.8, 3.0.6, 3.1.4, 3.2.2

PR #5340, opened 2023-04-10
Install openssl gem for Ruby 3.0.6

KostyaSha:
Related to https://github.com/rvm/rvm/pull/5309

* added missing file

### DIFF
--- a/gemsets/ruby/3.0.6/global.gems
+++ b/gemsets/ruby/3.0.6/global.gems
@@ -1,0 +1,1 @@
+openssl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 #### New features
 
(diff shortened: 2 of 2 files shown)


gemsets/ruby/3.0.6/global.gems
```diff
@@ -0,0 +1 @@
+openssl
```
pkuczynski, line 1: Can you add CHANGELOG entry please?

pkuczynski (approved):

gemsets/ruby/3.0.6/global.gems
KostyaSha: Btw, i don't see changelog entry for previous versions

KostyaSha:
It seems that https://github.com/rvm/rvm/pull/5336 also missing changelog entry

PR #5385, opened 2023-09-03
Fix truffleruby-23.0.0 installation on macos

jpcamara:
* When installing on an M1 macbook pro, the platform was coming through as macos, so the truffleruby checks did not identify a valid artifact id. That results in the url being invalid and the install failing
* Adding `macos-amd64` and `macos-aarch64` to the case statement with identical artifact ids to their `darwin` counterparts fixes the issue, and truffleruby installs correctly @eregon 

> Make sure that your pull request includes entry in the [CHANGELOG](CHANGELOG.md).

It's not clear to me where I would put an entry in the changelog, or how I would format it, but I am happy to add one! Also happy to add a spec, but it's unclear how i'd do that or what i'd test.

Thanks!

### DIFF
--- a/scripts/functions/selector_interpreters
+++ b/scripts/functions/selector_interpreters
@@ -280,6 +280,8 @@ __rvm_truffleruby_set_rvm_ruby_url()
       linux-aarch64)  truffleruby_artifact_id="FD40BA2367C226B6E0531518000AE71A" ;;
       darwin-amd64)   truffleruby_artifact_id="FD4AB182EA51EDFDE0531518000AF13E" ;;
       darwin-aarch64) truffleruby_artifact_id="FD40BBF6750C366CE0531518000ABEAF" ;;
+      macos-amd64)    truffleruby_artifact_id="FD4AB182EA51EDFDE0531518000AF13E" ;;
+      macos-aarch64)  truffleruby_artifact_id="FD40BBF6750C366CE0531518000ABEAF" ;;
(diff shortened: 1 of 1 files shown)


pkuczynski:
@eregon what do you think?

scripts/functions/selector_interpreters
```diff
@@ -280,6 +280,8 @@ __rvm_truffleruby_set_rvm_ruby_url()
       linux-aarch64)  truffleruby_artifact_id="FD40BA2367C226B6E0531518000AE71A" ;;
       darwin-amd64)   truffleruby_artifact_id="FD4AB182EA51EDFDE0531518000AF13E" ;;
       darwin-aarch64) truffleruby_artifact_id="FD40BBF6750C366CE0531518000ABEAF" ;;
```
eregon, line 282: Could you remove these two lines?
As on line 254 the platform is either `macos` or `linux`, never `darwin`

jpcamara: Sure thing - updated to have "macos" only

eregon:
It's unfortunate the CI didn't catch this, the test is skipped on macOS: https://github.com/rvm/rvm/blob/45caf951738afb34dd3c7f38b086166284dc5baa/.github/workflows/specs.yml#L31-L34
Might be worth a try to re-enable it.

eregon:
@pkuczynski It seems the failure on macOS CI is unrelated, could you merge this fix?

pkuczynski:
Sure, I can merge it

PR #5387, opened 2023-09-05
Fixing macOS CI and re-enable the truffleruby test on macOS

eregon:
It was disabled in https://github.com/rvm/rvm/pull/5119
and means this bug was not caught: https://github.com/rvm/rvm/pull/5385

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -18,16 +18,22 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: ./install
-      - run: source ~/.rvm/scripts/rvm && rvm use 2.7.7 --install --default && gem install tf -v '>=0.4.1'
-      - name: tests/fast/*
-        run: source ~/.rvm/scripts/rvm && tf --text tests/fast/*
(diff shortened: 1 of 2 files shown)


pkuczynski (dismissed):

eregon:
So test/fast/* passes now: https://github.com/rvm/rvm/actions/runs/8349160490/job/22852712175?pr=5387
But for some reason truffleruby_comment_test fails without any output.
I wonder if maybe the previous CI steps delete `~/.rvm/scripts/rvm` or cause that script to exit with 1.

eregon:
https://github.com/eregon/rvm/actions/runs/8358105456/job/22878671570
So `~/.rvm/scripts/rvm` exists but:
```
Run source ~/.rvm/scripts/rvm && true
Error: Process completed with exit code 1.
```
So it seems somehow the fast tests break rvm somehow, to the point it even fails to be sourced and the only error is the exit code, only on macOS (not on Linux).
I'll workaround by running the truffleruby tests first then, that should work fine.

pkuczynski (approved):

PR #5394, opened 2023-09-29
Update default openssl requirement to openssl@3

### LINKED ISSUE #3580: Error running '__rvm_make -j4' while running rvm install 2.2.3
I have tried uninstalling rvm but it did not work.

`ruby-2.2.3 - #removing src/ruby-2.2.3..
Searching for binary rubies, this might take some time.
No binary rubies available for: mint/17/i386/ruby-2.2.3.
Checking requirements for mint.
Requirements installation successful.
Installing Ruby from source to: /home/sushant/.rvm/rubies/ruby-2.2.3, this may take a while depending on your cpu(s)...
ruby-2.2.3 - #downloading ruby-2.2.3, this may take a while depending on your connection...
ruby-2.2.3 - #extracting ruby-2.2.3 to /home/sushant/.rvm/src/ruby-2.2.3....
ruby-2.2.3 - #configuring.........................................................
ruby-2.2.3 - #post-configuration..
ruby-2.2.3 - #compiling.......................................................................
Error running '__rvm_make -j4',
showing last 15 lines of /home/sushant/.rvm/log/1450313017_ruby-2.2.3/make.log
make[1]: *** [ext/openssl/all] Error 2
make[1]: *** Waiting for unfinished jobs....
installing default pty libraries
installing default date_core libraries
compiling psych_emitter.c
installing default psych libraries
linking shared-object pty.so
make[2]: Leaving directory `/home/sushant/.rvm/src/ruby-2.2.3/ext/pty'
linking shared-object psych.so
make[2]: Leaving directory `/home/sushant/.rvm/src/ruby-2.2.3/ext/psych'
linking shared-object date_core.so
make[2]: Leaving directory `/home/sushant/.rvm/src/ruby-2.2.3/ext/date'
make[1]: Leaving directory `/home/sushant/.rvm/src/ruby-2.2.3'
make: *** [build-ext] Error 2
++ return 2
There has been an error while running make. Halting the installation.`

This is what I have got after running `rvm info system`

`system:

  system:
    uname:       "Linux sushant-Inspiron-3421 3.13.0-24-generic #47-Ubuntu SMP Fri May 2 23:31:42 UTC 2014 i686 i686 i686 GNU/Linux"
    system:      "mint/17/i386"
    bash:        "/bin/bash => GNU bash, version 4.3.11(1)-release (i686-pc-linux-gnu)"
    zsh:         "/usr/bin/zsh => zsh 5.0.2 (i686-pc-linux-gnu)"
`

crpahl:
Fixes #3580.

* Update the default openssl version installed through Homebrew to `openssl@3`

The latest formulae in Homebrew have been [updated](https://github.com/Homebrew/homebrew-core/issues/134251) to use `openssl@3`. As a result, it's easy to get into a situation where multiple versions of openssl have been installed through brew. This will cause the rvm build to fail unless a specific version is specified with `--with-openssl-dir`. By changing the default, it aligns with the other formulae. After making this change I could do a straight forward:

`rvm install 3.2.2`


scripts/functions/requirements/osx_brew
```diff
@@ -331,7 +331,7 @@ requirements_osx_brew_define_openssl()
       ;;
 
     (*)
-      brew_openssl_package="openssl@1.1"
+      brew_openssl_package="openssl@3"
```
pkuczynski, line 334: This depends on the version of Ruby. Only latest can work with v3. Can you add a switch case detecting Ruby version and installing appropriate dependency?

pkuczynski (changes requested):

scripts/functions/requirements/osx_brew
crpahl: Ah, right. Yeah I can do that! I'm thinking the switch case should be updated so `openssl@3` is used for all Ruby versions (major and minor) going forward? There's probably a cleaner way to do this, but something like this for example (leaving the rest of the switch case as is)?

```
case "$1" in
  (ruby-3.[2-9]*|ruby-3.[1-9][0-9]*|ruby-[4-9]*)
    brew_openssl_package="openssl@3"
    ;;

  (ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
    brew_openssl_package="openssl"
    ;;

  (ree-1.8*)
    brew_openssl_package="openssl"
    ;;

  (rbx-1*|rbx-2*)
    brew_openssl_package="openssl"
    ;;

  (truffleruby*)
    brew_openssl_package="openssl"
    ;;

  (*)
    brew_openssl_package="openssl@1.1"
    ;;
esac
```

Or is there some way to clean up the switch case so `openssl@3` is the new default? This wouldn't be quite right, but just for example:

```
case "$1" in
    ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
        brew_openssl_package="openssl"
        ;;
    
    ree-1.8*)
        brew_openssl_package="openssl"
        ;;
    
    rbx-1*|rbx-2*)
        brew_openssl_package="openssl"
        ;;
    
    truffleruby*)
        brew_openssl_package="openssl"
        ;;
    
    ruby-3.0*|ruby-3.1*)
        brew_openssl_package="openssl@1.1"
        ;;
    
    *)
        brew_openssl_package="openssl@3"
        ;;
esac
```

Also, is using `openssl` for these other cases still correct? For example, I think this:

```
(ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
  brew_openssl_package="openssl"
  ;;
```

Would actually install `openssl@3` now? The new default has been updated [to be openssl@3](https://github.com/Homebrew/homebrew-core/blob/master/Aliases/openssl) I think? Sorry for the questions! Just working through some uncertainty.

jedrekdomanski:
@pkuczynski I was wondering when do you think this could go out to the master branch?

crpahl:
@pkuczynski What do you think of the change I just pushed?

- I think it addressed your feedback. Curious to know what you think now!
- `openssl` is _always_ given a specific version number. This will protect against the default formulae version changing again in the future (since we're not using the default)
- When/if a new version of openssl is added in the future, we just add a new case for it when that happens

scripts/functions/requirements/osx_brew
```diff
@@ -313,27 +313,31 @@ requirements_osx_brew_define_openssl()
...
+    brew_openssl_package="openssl@1.1"
+    ;;
+
+  truffleruby*)
+    brew_openssl_package="openssl@1.1"
+    ;;
+
+  ruby-3.0*|ruby-3.1*)
+    brew_openssl_package="openssl@1.1"
+    ;;
```
jcw-, line 335: both ruby 1 and 2 are covered in the first switch, to be consistent might make sense to also cover 3 there as well?

jcw- (commented):
makes sense to me - just one consistency comment

scripts/functions/requirements/osx_brew
```diff
@@ -313,27 +313,31 @@ requirements_osx_brew_define_openssl()
...
-    (truffleruby*)
-      brew_openssl_package="openssl"
-      ;;
-
-    (*)
-      brew_openssl_package="openssl@1.1"
-      ;;
-  esac
+case "\$1" in
+  ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
```
mikehardy, line 317:
```suggestion
  ruby-3.1*|ruby-3.0*|ruby-2.3*|ruby-2.2*|ruby-2.1*|ruby-2.0*|ruby-1.9*|ruby-1.8*)
```

...just hoping to move this along by providing commitable comments, cheers

scripts/functions/requirements/osx_brew
mikehardy, line 335:
```suggestion
```


...just hoping to move this along by providing commitable comments, cheers

scripts/functions/requirements/osx_brew
```diff
@@ -314,24 +314,24 @@ requirements_osx_brew_define_openssl()
...
-      brew_openssl_package="openssl"
+      brew_openssl_package="openssl@1.1"
       ;;
 
     (rbx-1*|rbx-2*)
-      brew_openssl_package="openssl"
+      brew_openssl_package="openssl@1.1"
       ;;
 
     (truffleruby*)
```
crpahl, line 329: truffleruby openssl support has also been updated recently and I'm wondering if this switch case also needs to be expanded for specific truffle ruby versions 🤔 

https://github.com/oracle/truffleruby/issues/2660#issuecomment-1158058741

jedrekdomanski:
@pkuczynski  When do you think this fix can be deployed? It's preventing me from installing any Ruby version.

phene:
Please update and merge this branch!

crpahl:
I'm going to close this PR. I actually think fully addressing this correctly should be left to an SME instead the more I learn about this area and the many edge cases across different Ruby versions and implementations.

We need a tool that currently doesn't have this problem so our dev department isn't blocked during onboarding and updates so we're going to use `asdf` for now. Thank you for the initial feedback though!

PR #5405, opened 2023-11-02
Add JRuby 9.3.13.0 and 9.4.5.0

headius:
Adds JRuby 9.3.13.0 and 9.4.5.0.


mathieujobin (approved):
not sure why CI is failing though...

pkuczynski:
What about 9.3.12? Seems like we are missing it?

pkuczynski:
> not sure why CI is failing though...

Probably openssl issue. It only fails on macOS.

headius:
The 9.3.12.0 release got a bit messed up and we had to skip it.

PR #5414, opened 2023-11-28
Added condition to check upstream-release if present

### LINKED ISSUE #3053: Add binary rubies for Linux Mint
In Linux Mint 17 I'm getting the error

```
No binary rubies available for: mint/17/x86_64/ruby-1.9.3-p547.
```

While I did eventually use `rvm mount` to install the ubuntu binary ruby, I feel it should be more straightforward:

See, every Mint release uses an Ubuntu base and, to the best of my knowledge, does not override anything Ruby related in its packages. That suggests that binary rubies for the corresponding Ubuntu base should be a fit for Mint.

My suggestion would be that rvm.io has rubies for Mint that are linked to their Ubuntu counterparts.
Of course, you might also opt to have rvm fall back to Ubuntu rubies programmatically.

Thank you for considering.

PS. While Linux Mint has up until now followed the Ubuntu release cycle version (15 was based on 13.04, 16 was based on 13.10), from now on Mint releases will be based on just the Ubuntu LTS releases (so 17, 18, 19 and 20 will all be based on 14.04)

scmanjarrez:
Fixes #3053 

- Check /etc/upstream-release/lsb_release if present in system that are based on another "base" distributions like Mint


scripts/functions/detect/system
```diff
@@ -21,7 +21,11 @@ __rvm_detect_system()
 
       _system_type="Linux"
 
-      if [[ -f /etc/lsb-release ]] && __rvm_detect_system_from_lsb_release
+      if [[ -f /etc/upstream-release/lsb-release ]] && __rvm_detect_system_from_lsb_release upstream
```
pkuczynski, line 24: I think we should implement it the other way around. First check the `lsb_release` and only when it returns `1` (fails), check the `upstream-release`.

scripts/functions/detect/system_name/lsb_release
```diff
@@ -2,7 +2,11 @@
 
 __rvm_detect_system_from_lsb_release()
 {
-  local __system_name="$( awk -F'=' '$1=="DISTRIB_ID"{print $2}' /etc/lsb-release | head -n 1 | tr '[A-Z]' '[a-z]' | tr -d \" )"
+  local __lsb_release_file=""
+  if [ -n "$1" ]; then
```
pkuczynski, line 6: We should be more explicit here and compare against `upstream`.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Add Support for OpenSSL 1.1/3.0 to older ruby versions [\#5248](https://github.com/rvm/rvm/pull/5248)
 * Update truffleruby-dev to use builds based on Ubuntu 20.04 [\#5322](https://github.com/rvm/rvm/issues/5322)
 * Install openssl gem for Ruby 3.0.6 [\#5340](https://github.com/rvm/rvm/pull/5340)
+* Patch to take into account /etc/upstream-release/lsb-release if present when detecting available binaries [\#3053](https://github.com/rvm/rvm/issues/3053)
```
pkuczynski, line 11:
```suggestion
* Consider `upstream-release` to detect operating system [\#3053](https://github.com/rvm/rvm/issues/3053)
```

pkuczynski (changes requested):
Thank you for the contribution. Left some comments. Afterwards we can merge it I think.

scripts/functions/detect/system
scmanjarrez: The problem with that is lsb_release is also present in mint, so it would enter that check returning incorrect value.

pkuczynski: I see your point. If `mint` follows ubuntu, maybe we should just delete `requirements/mint` and remove it from `lsb_release/__rvm_detect_system_from_lsb_release`? This way `__rvm_detect_system()` would return `1` and the proposed check on `upstream-release` would work correctly.

scmanjarrez: Hi, sorry for late response, I was out of home. 
Mmmm, I don't know the implications of doing that, I don't know the code of rvm very well (I'm new to rvm code)

pkuczynski (approved):

PR #5425, opened 2023-12-28
Fix incorrect example for rvm_make_flags (string changed to array)

darkBuddha:
- fixes incorrect example (`-j` make flag must be provided as array, not as string)
- added another example

cf. https://stackoverflow.com/a/12878459/4490968


pkuczynski (approved):

pkuczynski:
Thanks! And sorry for delayed merge

PR #5456, opened 2024-04-26
Upgrade github actions and truffleruby version in CI

torrocus:
## Pull Request Summary

I changed the `actions/checkout` version because it was outdated.
I also noticed that we still had a problem with failing tests, so I tried to fix it.

## Feedback

If the proposed change with TruffleRuby versions is not ok, I can remove it.

Updated: 
CI testing goes further. Now they fail on `tests/fast/do_comment_test.sh`

### DIFF
--- a/tests/long/truffleruby_comment_test.sh
+++ b/tests/long/truffleruby_comment_test.sh
@@ -17,6 +17,6 @@ rvm truffleruby-head do ruby -v # status=0; match=/truffleruby/
 rvm remove truffleruby-head # status=0
 
 ## Test that the right version is installed (#4633)
-rvm install truffleruby-1.0.0-rc13 # status=0; match!=/Already installed/
-rvm truffleruby-1.0.0-rc13 do ruby -v # status=0; match=/truffleruby 1.0.0-rc13/
-rvm remove truffleruby-1.0.0-rc13 # status=0
+rvm install truffleruby-23.1.2 # status=0; match!=/Already installed/
+rvm truffleruby-23.1.2 do ruby -v # status=0; match=/truffleruby 23.1.2/
+rvm remove truffleruby-23.1.2 # status=0
--- a/.github/workflows/gpg_signature.yml
+++ b/.github/workflows/gpg_signature.yml
@@ -7,7 +7,7 @@ jobs:
   signature-check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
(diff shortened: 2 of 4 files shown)


tests/long/truffleruby_comment_test.sh
```diff
@@ -17,6 +17,6 @@ rvm truffleruby-head do ruby -v # status=0; match=/truffleruby/
 rvm remove truffleruby-head # status=0
 
 ## Test that the right version is installed (#4633)
-rvm install truffleruby-1.0.0-rc13 # status=0; match!=/Already installed/
-rvm truffleruby-1.0.0-rc13 do ruby -v # status=0; match=/truffleruby 1.0.0-rc13/
-rvm remove truffleruby-1.0.0-rc13 # status=0
```
torrocus, line 22: Here's a snippet of the logs I can see locally:
```
Recompiling the OpenSSL C extension (libssl version is incompatible)
The clang, opt and llvm-link tools, part of LLVM, do not appear to be available in $PATH.
```

eregon: Yeah updating the truffleruby version here seems good.
(I think it was working fine in CI, so it's probably expected some tests only pass on CI.)

eregon (approved):
Any idea why the macOS CI already fails again? 😢

tests/long/truffleruby_comment_test.sh
eregon: Ah it was failing in https://github.com/rvm/rvm/actions/runs/8845813076/job/24290469426 I see, maybe the side effect of `macos-latest` now using a newer macOS version, maybe we should pin to a given macos-NN version to avoid surprises like this.

eregon: Yep, looks like it, that job ran on 14.4 and the last green job https://github.com/rvm/rvm/actions/runs/8713093296/job/23900815353#step:1:4 ran on 12.7.
I think we should pin to `macos-12` for now to fix the macos CI, and that should hopefully fix macos CI.
We can still update the TruffleRuby version tested it, that's fine.

torrocus: Unfortunately, I don't know what could have happened. My change exposed the problem elsewhere. If we want to solve it properly, we have to go through the full way.

eregon: Looking at these two logs, it seems `macos-latest` used to be macos 12.7 arm64 and now resolved to macos 14 aarch64. Pretty weird stuff.

PR #5458, opened 2024-04-28
Parallelize GitHub Action jobs in CI

torrocus:
## Pull Request Summary

CI tests take a long time because they run different cases one after the other. If something fails, no further tests will be run. ~CI tests have not been passed for some time now. Fast tests didn't pass on CI for MacOS. This change is intended to help us solve this problem more easily.~

* split a long-running tests job into smaller jobs that can be parallelized

## Feedback

The total CI execution time is the same, but the waiting time for the test results is shorter.

### Before

Screenshot before

### After

Screenshot after

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -11,6 +11,20 @@ jobs:
         os:
           - ubuntu-latest
           - macos-13
+        test:
+          # For some reason, after running these tests,
+          # `source ~/.rvm/scripts/rvm` fails on macOS, so run them last.
+          # See https://github.com/rvm/rvm/pull/5387#issuecomment-2009391015
+          # These tests also change the default ruby, which is another reason to run them last.
+          - "tests/fast/*"
+          - "tests/long/truffleruby_comment_test.sh"
+        include:
+          - os: ubuntu-latest
+            # works on local, but fails in CI, needs to be investigated
+            test: "tests/long/named_ruby_and_gemsets_comment_test.sh"
+          - os: ubuntu-latest
+            # https://github.com/rvm/rvm/issues/4937
+            test: "tests/long/ruby_prepare_mount_comment_test.sh"
     runs-on: ${{ matrix.os }}
     env:
       TERM: ansi
@@ -20,20 +34,4 @@ jobs:
       - run: ./install
       - run: source ~/.rvm/scripts/rvm && rvm use 2.7.7 --install --default
       - run: source ~/.rvm/scripts/rvm && gem install tf -v '>=0.4.1'
-
-      - name: named_ruby_and_gemsets_comment_test
-        if: ${{ !contains(matrix.os, 'macos') }} # works on local, but fails in CI, needs to be investigated
-        run: source ~/.rvm/scripts/rvm && tf --text tests/long/named_ruby_and_gemsets_comment_test.sh
-
-      - name: ruby_prepare_mount_comment_test
-        if: ${{ !contains(matrix.os, 'macos') }} # https://github.com/rvm/rvm/issues/4937
-        run: source ~/.rvm/scripts/rvm && tf --text tests/long/ruby_prepare_mount_comment_test.sh
-
-      - name: truffleruby_comment_test
-        run: source ~/.rvm/scripts/rvm && tf --text tests/long/truffleruby_comment_test.sh
-
-      # For some reason, after running these tests, `source ~/.rvm/scripts/rvm` fails on macOS, so run them last.
-      # See https://github.com/rvm/rvm/pull/5387#issuecomment-2009391015
-      # These tests also change the default ruby, which is another reason to run them last.
-      - name: tests/fast/*
-        run: source ~/.rvm/scripts/rvm && tf --text tests/fast/*
+      - run: source ~/.rvm/scripts/rvm && tf --text ${{ matrix.test }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,7 @@
 
 * Migrate from rvm_io.global.ssl.fastly.net to rvm-io.global.ssl.fastly.net to fix domain fronting report from Fastly [\#5438](https://github.com/rvm/rvm/pull/5438)
 * Upgrade github actions and truffleruby version in CI [\#5456](https://github.com/rvm/rvm/pull/5456)
+* Parallelize GitHub Action jobs in CI [\#5458](https://github.com/rvm/rvm/pull/5458)
 
 ## [1.29.12](https://github.com/rvm/rvm/releases/tag/1.29.12)
 15 January 2021 - [Full Changelog](https://github.com/rvm/rvm/compare/1.29.11...1.29.12)


torrocus:
@pkuczynski what do you think about this change?

.github/workflows/specs.yml
```diff
@@ -22,18 +48,42 @@ jobs:
...
+          - ubuntu-latest
+          - macos-latest
+    runs-on: ${{ matrix.os }}
+    env:
+      TERM: ansi
+      RVM_SKIP_BREW_UPDATE: true
+    steps:
+      - uses: actions/checkout@v4
+      - run: ./install
+      - run: source ~/.rvm/scripts/rvm && rvm use 2.7.7 --install --default
```
pkuczynski, line 85: Isnt this one for trufflefurby?

pkuczynski:
It make sense to run things in parallel if it speeds up the tests, that's for sure. Just the code duplication is aweful. Can we somehow avoid this?

@eregon wdyt?

eregon:
What about using a 2-dimensional matrix for that? That should address the duplication and keep things concise.

eregon:
> CI tests have not been passed for some time now.

Only on macOS which is a PITA. They work fine on Linux AFAIK.

.github/workflows/specs.yml
```diff
@@ -3,15 +3,41 @@ on: [push, pull_request]
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
eregon, line 8: I don't think we should cancel CI jobs

torrocus: Let me explain it. Canceling is only when we push new changes to the same branch. This allows you to save running time on GitHub if you make frequent changes. This does not affect CI if work is carried out on different branches.

eregon: I know what it does, but for instance it could cancel a CI run on master if 2 PRs are merged quickly one after another, and that seems bad (could make it harder to find a faulty change) + there is not really enough CI load on this repo to need to cancel anything.

torrocus: We can exclude cancel for branch master. WDYT?

eregon: But then that prevents to run CI jobs on different commits on master in parallel, which is unnecessary. Please remove this section entirely, there is no need to change this.

.github/workflows/specs.yml
torrocus: I don't want to guess, so let's check it out.

torrocus: The [tests didn't pass](https://github.com/rvm/rvm/actions/runs/8940912966), so no.

torrocus:
> @pkuczynski said:
> It make sense to run things in parallel if it speeds up the tests, that's for sure. Just the code duplication is aweful. Can we somehow avoid this?

I don't like these duplications either. Unfortunately, every job must perform similar activities:
- git checkout repo
- install ruby via rvm
- install tf gem
- run selected tests

At the moment I have no idea how to do it better, although I would love to share these parts of the job.


> @eregon said:
> What about using a 2-dimensional matrix for that? That should address the duplication and keep things concise.

If you mean strategies (matrix for OS), it won't change anything. To introduce parallelism, we need separate jobs. Anyway, where we run jobs on different OSs, I used matrix. I only removed where it was conditional.

eregon:
Some general feedback: I think 9min vs 14min doesn't make a huge difference, in either case no one will wait for that actively. And then of course it will redo some work multiple times which is rather wasteful (e.g. compile ruby 2.7.7).
But having isolation between these groups of tests is valuable, especially given the existing issues between them, so this seems worth it.

eregon:
> If you mean strategies (matrix for OS), it won't change anything. To introduce parallelism, we need separate jobs. Anyway, where we run jobs on different OSs, I used matrix. I only removed where it was conditional.

Use a matrix e.g. like here: https://github.com/ruby/spec/blob/1afa2eab1cab5df6c5706fc5b7e92756b1db4f07/.github/workflows/ci.yml#L10-L12
Those jobs all run in parallel.
And that solves the duplication, at the cost of some existing `if` conditionals for tests which don't work on macOS yet (but should eventually).

You can even avoid those `if:` with an explicit include list like https://github.com/ruby/spec/blob/1afa2eab1cab5df6c5706fc5b7e92756b1db4f07/.github/workflows/ci.yml#L14-L17

.github/workflows/specs.yml
eregon: I think it's more the opposite: `truffleruby_comment_test.sh` might not need `2.7.7` to be installed. But it feels something out of scope and might not hold either, so I wouldn't change that in this PR.

torrocus:
> Some general feedback: I think 9min vs 14min doesn't make a huge difference, in either case no one will wait for that actively. And then of course it will redo some work multiple times which is rather wasteful (e.g. compile ruby 2.7.7). But having isolation between these groups of tests is valuable, especially given the existing issues between them, so this seems worth it.

Yes, separation is a big advantage. If we add caching in the future, we should gain more time.

torrocus:
> > If you mean strategies (matrix for OS), it won't change anything. To introduce parallelism, we need separate jobs. Anyway, where we run jobs on different OSs, I used matrix. I only removed where it was conditional.
> 
> Use a matrix e.g. like here: https://github.com/ruby/spec/blob/1afa2eab1cab5df6c5706fc5b7e92756b1db4f07/.github/workflows/ci.yml#L10-L12 (the keys would be `os` and `test` instead) Those jobs all run in parallel. And that solves the duplication, at the cost of some existing `if` conditionals for tests which don't work on macOS yet (but should eventually).
> 
> You can even avoid those `if:` with an explicit include list like https://github.com/ruby/spec/blob/1afa2eab1cab5df6c5706fc5b7e92756b1db4f07/.github/workflows/ci.yml#L14-L17

Interesting idea. I'll see what I can do with it.

torrocus:
@pkuczynski @eregon I made changes and it looks better. What do you think?

.github/workflows/specs.yml
```diff
@@ -3,14 +3,26 @@ on: [push, pull_request]
...
           - ubuntu-latest
           - macos-13
+        test:
+          - "tests/fast/*"
+          - "tests/long/truffleruby_comment_test.sh"
+        include:
+          - os: ubuntu-latest
+            test: "tests/long/named_ruby_and_gemsets_comment_test.sh"
+          - os: ubuntu-latest
+            test: "tests/long/ruby_prepare_mount_comment_test.sh"
```
eregon, line 21: Could you preserve the comments for these 2 jobs (as to why they are not run on macOS)?

.github/workflows/specs.yml
```diff
@@ -3,14 +3,26 @@ on: [push, pull_request]
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref_name != 'master' }}
+
```
eregon: Could you remove this?

.github/workflows/specs.yml
torrocus: I feel like removing will hurt the project, but I'll adapt.

eregon (commented):
Looks great now, 2 more comments

.github/workflows/specs.yml
eregon: Because CI would start with a delay? I've never observed it on rvm/rvm, so I think it's a not an issue in practice (the default numbers of runners is pretty high).
And for the scenario I mentioned for master it's actually worse, adding a constraint which is not needed.
Another downside is e.g. when working on CI one might push more changes but still be interested in older in-progress CI jobs (e.g. I still want to see macOS CI results when touching common logic). This would cancel them and make it very annoying for those cases, which IMO outweighs any possible gain from this.

eregon (approved):

.github/workflows/specs.yml
torrocus: The way I see it is that if I make changes, I would like to minimize the time in which I receive information that the tests have passed and I don't want to wait until the entire CI queue is completed. I completely agree that the master branch should be unique in this case. But maybe it won't be such a big problem. If it hurts, we can consider the proposed changes.

.github/workflows/specs.yml
```diff
@@ -11,6 +11,20 @@ jobs:
         os:
           - ubuntu-latest
           - macos-13
+        test:
+          - "tests/fast/*"
+          # For some reason, after running these tests,
+          # `source ~/.rvm/scripts/rvm` fails on macOS, so run them last.
+          # See https://github.com/rvm/rvm/pull/5387#issuecomment-2009391015
+          # These tests also change the default ruby, which is another reason to run them last.
```
eregon, line 19: This comment applies to `tests/fast/*`, not `tests/long/truffleruby_comment_test.sh`.
But I think best to remove it, it's not relevant anymore since they run in separate jobs.

eregon (dismissed):

.github/workflows/specs.yml
torrocus: Right, I missed that. Well caught.

eregon (approved):

eregon:
@pkuczynski Could you merge? (I cannot)
And also adapt the Required status checks to the new job names?

torrocus:
@pkuczynski Since we have changed the CI process, we also need to change the settings in -> Settings -> Branches -> Branch protection rules. Then select Edit for the "master" branch. And then in the "Require status checks to pass before merging" option, remove the status checks "tests" and add "CI/tests".
Otherwise, only people with force push permission can merge ("tests" check status is waiting for report over and over again).

pkuczynski:
Sure I can :) Give me few mins.

pkuczynski (approved):

pkuczynski:
Thank you @torrocus for your contribution

eregon:
> MacOS tests are taking ridicilous long. Maybe we could check why...

Looking at https://github.com/rvm/rvm/actions/runs/8945080363/usage and specific logs for test/fast/*, it looks like on Linux it's mostly (only?) prebuilt Ruby binaries being used, while on macOS every ruby installed seems compiled from source, which takes a while.

One thing to improve might be to always use the same Ruby version for all tests and avoid uninstalling it.
And obviously if there could be prebuilt Rubies on macOS that'd be even better.

pkuczynski:
That make sense. We had issues with building movable ruby in the past for macOS :(

PR #5470, opened 2024-05-30
Add support for Ruby 3.3.2 (#5470)

moritori:

    Add support for Ruby 3.3.2


havenwood (changes requested):
Thank you for the PR!

These additions look perfect. Since 3.3.2 is now latest stable, we should also bump `ruby_version=3.3.1` to `ruby_version=3.3.2` in `config/db`.

moritori:
I missed that. Thanks for pointing it out. I've just pushed the correction.

havenwood (approved):

PR #5477, opened 2024-06-15
Add support for unversioned libncurses-dev packages on newer Debian-like systems

### LINKED ISSUE #5475: libncurses5-dev dependency broken on Ubuntu 24.04
On Ubuntu 24.04, rvm wants to install `libncurses5-dev`. That package, however, [does not exist in the `noble` repos](https://packages.ubuntu.com/noble/libncurses5-dev), and you get redirected to [`libncurses-dev`, which installs version 6.4](https://packages.ubuntu.com/noble/libncurses-dev). The RVM requirement check does not accept that, though.

I monkey-patched `scripts/functions/requirements/debian` to use the version 6 package, and building Ruby 3.3.3 worked just fine.

denschub:
I tested this on Ubuntu Noble, Ubuntu Focal, Debian Bookworm, Debian testing (future Trixie), and Debian unstable, and I could install Ruby 3.3 on all systems. I couldn't test all the other distributions, but this shouldn't break anything that wasn't broken before.

This technically doesn't work on Debian testing/unstable, as the numeric version number detection doesn't work, and instead it just discovers the version number as `trixie_sid`, but the old package is still aliased, so installing it works just fine. This will need adjustment when Trixie is released, but that's not going to happen tomorrow.

This fixes #5475.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * Fix truffleruby-head installation on macos-aarch64 [\#5446](https://github.com/rvm/rvm/pull/5446)
 * Fix bug with Bootsnap on Ruby 3.3.1 [\#5457](https://github.com/rvm/rvm/pull/5457)
 * Use pkgconf instead of pkg-config in Solus [\#5471](https://github.com/rvm/rvm/pull/5471)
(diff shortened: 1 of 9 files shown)


pkuczynski (approved):

pkuczynski:
Great work @denschub! Thanks!

PR #5538, opened 2025-01-21
Add JRuby 9.4.10.0

headius:
* Adds JRuby 9.4.10.0


headius:
I don't think any of my changes would have caused these failures.

PR #5541, opened 2025-01-29
Add JRuby 9.4.11.0

headius:
Add JRuby 9.4.11.0.


headius:
@pkuczynski I followed your lead on labels and milestone but just noticed the signature requirement. I'll try to get my commits signing again before the next release.

PR #5588, opened 2025-09-21
Add Ruby 3.4.6

kenhys:
See https://www.ruby-lang.org/en/news/2025/09/16/ruby-3-4-6-released/


kenhys:
macos CI fails, but it seems that it was not introduced by this PR.

Issue #5615, opened 2025-12-29
Can not compile Ruby 4 as movable


lvangool:
When compiling Ruby 4 with the --movable flag I get the error:
```
Only MRI Ruby 1.9.3+ can be compiled movable with RVM
```
This looks like a missing matcher, PR incoming!

PR #5616, opened 2025-12-29
Add ruby-4* matchers for compiling movable Ruby

lvangool:
Fixes https://github.com/rvm/rvm/issues/5615

Recognise Ruby 4.0.0 as a valid target for movable compilation


pkuczynski:
Thanks!

PR #5618, opened 2025-12-30
ci: enable fast-tests for Ruby 4.0

kenhys:
Followup of #5611

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -22,6 +22,7 @@ jobs:
           - 3.2
           - 3.3
           - 3.4
+          - '4.0'
         include:
           # TODO: works on local, but fails in CI, needs to be investigated
           - os: ubuntu-latest


.github/workflows/specs.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           - 3.2
           - 3.3
           - 3.4
+          - '4.0'
```
pkuczynski:
```suggestion
          - 4.0
```

pkuczynski (approved):

pkuczynski:
We also have some tests which works on local but fail in CI like [these](https://github.com/rvm/rvm/blob/master/.github/workflows/specs.yml#L27) and some not working in newer riuby like [these](https://github.com/rvm/rvm/blob/master/.github/workflows/specs.yml#L47). Would you like to look at those?

PR #5626, opened 2026-01-21
Add TruffleRuby 33.0.1

eregon:
https://github.com/truffleruby/truffleruby/releases/tag/graal-33.0.1

### DIFF
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -1,4 +1,5 @@
 name: Test
+description: Test
 
 inputs:
   ruby:
@@ -22,3 +23,5 @@ runs:
 
     - run: source ~/.rvm/scripts/rvm && tf --text ${{ inputs.test }}
       shell: bash
+      env:
+        TF_DEBUG: 1


eregon:
Found TF_DEBUG=1.
The output on Linux https://github.com/rvm/rvm/actions/runs/21243344041/job/61126362720?pr=5626#step:5:69
looks a bit weird because the download output is before `##### starting test truffleruby.` which does not make sense timing-wise.
Maybe `tf` or RVM (or GitHub Actions) is buffering/delaying stdout but not stderr?

On macos-latest it looks stuck just after the download like in in the previous runs on this PR:
https://github.com/rvm/rvm/actions/runs/21243344041/job/61126362729?pr=5626
Not printing any command that tf runs nor even `##### starting test truffleruby.`.

eregon:
> Maybe `tf` or RVM (or GitHub Actions) is buffering/capturing/delaying stdout but not stderr?

`tf` must capture output for tests like
```
rvm install truffleruby # status=0; match!=/Already installed/; match=/compiling c-extensions/
```
But ideally it'd print a line as soon as it receives it or so.
(TruffleRuby itself doesn't do write buffering so that should be no issue.)

Not quite sure how to debug this.

PR #5652, opened 2026-05-09
Fix nounset loading errors

### LINKED ISSUE #4694: /usr/local/rvm/scripts/functions/support: line 182: _system_name: unbound variable
## Description

RVM suddenly started throwing an "unbound variable" error while trying to install and use Ruby 2.3.4 on Amazon Linux.  We first noticed the problem on May 8, 2019 and I believe it might be related to the latest RVM release.

## Steps to reproduce

We are using Packer to build AWS images with RVM and Ruby 2.3.4 installed.  Our process runs from a Jenkins job and has been working fine every Sunday for the past couple of years.  During the course of building a new Jenkins environment last week we discovered that the Packer process was failing with a `_system_name: unbound variable` error and we thought it was something that we had done wrong.  Investigating a bit more, we discovered that our latest Sunday job had failed in the same way.

 1. Configure Packer on AWS
 1. Install the latest RVM (on Packer on AWS)
 1. Install Ruby 2.3.4 with openSSL (`yes | rvm install ruby-2.3.4 -- --with-openssl-dir=/etc/pki/tls`
 1. Use the newly installed Ruby 2.3.4
 1. Get an unbound variable error 

## Expected behavior
We expected to see Ruby install and our Packer process continue as normal.

## Actual behavior
The actual behavior was an error that we are able to replicate using 2 different AWS accounts and Jenkins instances.

## Gist
Here's a Gist that shows the bash script that we are running inside Packer and the output from Jenkins.  The Jenkins DEBUG flag is turned on and shows the complete stack dump from RVM.

dicnunz:
Fixes #4694.

IssueHunt bounty: https://oss.issuehunt.io/r/rvm/rvm/issues/4694 ($40)

The reported script runs under `set -euo pipefail`, so ordinary RVM loading must not read unset shell variables. I reproduced a current nounset failure locally with:

```sh
bash -u -c 'export rvm_path=$PWD; source scripts/rvm >/tmp/rvm-source-nounset.out; rvm current'
```

This currently fails before `rvm current` because `rvm_prefix` is read without a default. After fixing that, the same nounset path exposed the `_system_name` case in the DB/system setup path, so this also makes the system-key DB lookup tolerate unset detection variables and fall back to the generic key.

Changes:

- Guard `rvm_prefix` while guessing the prefix during `scripts/rvm` load.
- Guard early utility setup before `_system_name` has been detected.
- Make `__rvm_db_system` build system-specific keys only when the relevant system variables are present, then fall back to the generic key.
- Add a fast nounset regression test for both `source scripts/rvm` and the `_system_name` DB fallback path.
- Add changelog entry.

Tests:

- `env rvm_path=$PWD BUNDLE_GEMFILE=$PWD/tests/Gemfile BUNDLE_PATH=/tmp/rvm-tf-bundle bundle exec tf --text tests/fast/nounset_comment_test.sh`
- `bash -u -c 'unset GEM_HOME GEM_PATH MY_RUBY_HOME; export rvm_path=$PWD; source scripts/rvm >/dev/null; rvm current'`
- `bash -u -c 'export rvm_path=$PWD rvm_scripts_path=$PWD/scripts rvm_user_path=$PWD/user; source scripts/base >/dev/null; unset _system_name _system_version; __rvm_db_system ruby_version'`
- `bash -n scripts/rvm scripts/functions/db scripts/functions/support tests/fast/nounset_comment_test.sh`
- `git diff --check -- CHANGELOG.md scripts/rvm scripts/functions/db scripts/functions/support tests/fast/nounset_comment_test.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 #### Bug fixes
 
+* Fix nounset errors while loading RVM with `set -u` [\#4694](https://github.com/rvm/rvm/issues/4694)
 * Remove unsupported libclang, libclang-dev, openssl-dev and zlib-dev from Termux requirements, and fix installation by not calling getent [\#5101](https://github.com/rvm/rvm/pull/5101)
 * Allow ruby-head to be installed using ruby 3, not truffleruby [\#5115](https://github.com/rvm/rvm/pull/5115)
 * Fix attempts to install/uninstall ruby picking jruby, mruby or truffleruby instead [\#5116](https://github.com/rvm/rvm/pull/5116)
--- a/scripts/functions/db
+++ b/scripts/functions/db
@@ -5,7 +5,18 @@
 __rvm_db_system()
 {
   \typeset __key __message
-  for __key in "${_system_name}_${_system_version}_$1" "${_system_name}_$1" "$1"
+  \typeset -a __keys
+
+  __keys=()
+  if [[ -n "${_system_name:-}" && -n "${_system_version:-}" ]]
+  then __keys+=( "${_system_name}_${_system_version}_$1" )
+  fi
+  if [[ -n "${_system_name:-}" ]]
+  then __keys+=( "${_system_name}_$1" )
+  fi
+  __keys+=( "$1" )
+
+  for __key in "${__keys[@]}"
   do
     if __rvm_db "${__key}_error" __message
     then rvm_error "${__message}"
@@ -13,7 +24,7 @@ __rvm_db_system()
     if __rvm_db "${__key}_warn" __message
     then rvm_warn "${__message}"
     fi
-    if __rvm_db "${__key}" "$2"
+    if __rvm_db "${__key}" "${2:-}"
     then return 0
     fi
   done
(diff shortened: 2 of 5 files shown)


dicnunz:
Rebased this branch on current master and revalidated the nounset fix at `6223f2e8`.

Proof:
- `BUNDLE_GEMFILE=tests/Gemfile BUNDLE_PATH=/tmp/rvm-tf-bundle bundle exec tf --text tests/fast/nounset_comment_test.sh` -> 2 commands / 2 assertions passed
- `bash -u -c 'unset GEM_HOME GEM_PATH MY_RUBY_HOME; export rvm_path=$PWD; source scripts/rvm >/dev/null; rvm current'` -> `system`
- `bash -u -c 'export rvm_path=$PWD rvm_scripts_path=$PWD/scripts rvm_user_path=$PWD/user; source scripts/base >/dev/null; unset _system_name _system_version; __rvm_db_system ruby_version'` -> current ruby version
- `bash -n scripts/rvm scripts/functions/db scripts/functions/support tests/fast/nounset_comment_test.sh`
- `git diff --check upstream/master..HEAD`

The only remote Mergeable failure is the repository label rule: `1/8 Fail(s): LABEL`, because the PR has no label. I tried adding `bug` through the API, but GitHub returned `403 Must have admin rights to Repository`. A maintainer label/re-review should be the remaining external gate.

